PR #9, opened 2025-09-02
added connection edit functionality and extended the sidebar with new…

nassery318:
… options

### DIFF
--- a/apps/frontend/src/components/Dashboard.tsx
+++ b/apps/frontend/src/components/Dashboard.tsx
@@ -3,6 +3,8 @@ import { selectData } from "@/state/valkey-features/info/infoSelectors.ts";
 import { Card } from "./ui/card";
 import { AppHeader } from "./ui/app-header";
 import { LayoutDashboard } from "lucide-react";
+import { selectConnectionDetails } from "@/state/valkey-features/connection/connectionSelectors";
+
 
 export function Dashboard() {
   const {
@@ -11,16 +13,16 @@ export function Dashboard() {
     connected_clients,
     keys_count,
     bytes_per_key,
-    tcp_port,
     server_name,
   } = useSelector(selectData);
+  const connectionDetails = useSelector(selectConnectionDetails);
   return (
     <div className="p-4">
       <AppHeader
         title="Dashboard"
         icon={<LayoutDashboard size={20} />}
         servername={server_name || ""}
-        port={tcp_port || ""}
+        port={connectionDetails.port}
       />
       <div className="flex flex-wrap gap-4">
         {[
--- a/apps/frontend/src/components/ui/app-sidebar.tsx
+++ b/apps/frontend/src/components/ui/app-sidebar.tsx
@@ -5,6 +5,9 @@ import {
   ChartNoAxesCombined,
   ChevronLeft,
   ChevronRight,
+  Cog,
+  CircleQuestionMark,
+  Github,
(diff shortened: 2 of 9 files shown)


apps/frontend/src/components/Dashboard.tsx
```diff
@@ -3,6 +3,8 @@ import { selectData } from "@/state/valkey-features/info/infoSelectors.ts";
 import { Card } from "./ui/card";
 import { AppHeader } from "./ui/app-header";
 import { LayoutDashboard } from "lucide-react";
+import { selectConnectionDetails } from "@/state/valkey-features/connection/connectionSelectors";
+
```
arseny-kostenko, line 7: extra empty line

apps/frontend/src/components/ui/app-sidebar.tsx
```diff
@@ -98,6 +89,44 @@ export function AppSidebar() {
...
         {isExpanded ? <ChevronLeft size={15} /> : <ChevronRight size={15} />}
       </button>
+      <div>
+        <ul className="space-y-2">
+          {[
+            {
+              href: "https://github.com/",
+              title: "GitHub",
+              icon: Github,
+              isExternal: true,
```
arseny-kostenko, line 99: is it basically href.startsWith("https://") or isInternal === href.startsWith("/") ?

apps/frontend/src/components/ui/edit-form.tsx
```diff
@@ -0,0 +1,133 @@
...
+        username,
+        password,
+      })
+    );
+
+    // Disconnect First
+    dispatch(valkeySetConnected(false));
+
+    // Then Reconnect
+    setTimeout(() => {
```
arseny-kostenko, line 52: this has to be in the epic

@ravjotbrar, if your PR is still pending — you can rebase/merge master once this is merged and refactor it in your branch

ravjotbrar: It's not pending, but I will refactor in my next PR if I have time.

apps/frontend/src/components/ui/app-sidebar.tsx
nassery318: isExternal flag is for the <a> tag which directs user to external link of GitHub, if there is no isExternal flag it is using the <Link> which will route user internally to the settings page.

arseny-kostenko: yeah, I mean, you set it manually, and your logic for this was implicit, it's totally in your head — if somebody else has to maintain or extend this code, they will be looking at it and having to come up with their own logic of determining if a link isExternal, and we should just hope that they know that external probably means that it's not served by our app, AND *currently* if it's external it should be `<a>` instead of `<Link>` — there is a lot of implicit assumptions.

so to make this code entirely clear for future generations by making implicit — explicit, you have two choices:
1. Leave a comment explaining why this property exists and how url => isExternal? 
2. Make it an implementation detail, remove the property from the array and instead of `item.isExternal ?` do `item.href.startsWith("https://") ?` or flip it to isInternal to be protocol agnostic `item.href.startsWith("/") ?`

this is also a not critical comments, so it's not blocking the merge. I just thought it's worth pointing out.

nassery318: Got the point, will be fixing in the next PR

PR #10, opened 2025-09-02
Change connection slices

ravjotbrar:
Change connection slices to use connectFulfilled, connectPending, and connectRejected. Also, added proper disconnect functionality and refactored epics.

### DIFF
--- a/apps/frontend/src/components/Connection.tsx
+++ b/apps/frontend/src/components/Connection.tsx
@@ -14,6 +14,7 @@ import { resetConnection } from "@/state/valkey-features/connection/connectionSl
 import { HousePlug, Unplug, Pencil } from "lucide-react";
 import ConnectionForm from "./ui/connection-form";
 import EditForm from "./ui/edit-form";
+import { CONNECTED } from "@common/src/constants"
 
 export function Connection() {
   const dispatch = useAppDispatch();
@@ -24,7 +25,7 @@ export function Connection() {
   const connectionDetails = useSelector(selectConnectionDetails);
   const { server_name } = useSelector(selectData);
 
-  const isConnected = useSelector(selectStatus) === "Connected";
+  const isConnected = useSelector(selectStatus) === CONNECTED
   const hasRedirected = useSelector(selectRedirected);
 
   const handleDisconnect = () => {
--- a/apps/frontend/src/components/ui/edit-form.tsx
+++ b/apps/frontend/src/components/ui/edit-form.tsx
@@ -7,9 +7,9 @@ type EditFormProps = {
 
 import {
   updateConnectionDetails,
-  setConnecting as valkeySetConnecting,
+  connectPending,
 } from "@/state/valkey-features/connection/connectionSlice.ts";
-import { setConnected as valkeySetConnected } from "@/state/valkey-features/connection/connectionSlice.ts";
+import { resetConnection } from "@/state/valkey-features/connection/connectionSlice.ts";
 import { selectConnectionDetails } from "@/state/valkey-features/connection/connectionSelectors";
 import { useState, useEffect } from "react";
 import { useSelector } from "react-redux";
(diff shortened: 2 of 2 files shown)


apps/frontend/src/components/ui/app-sidebar.tsx
```diff
@@ -25,14 +25,12 @@ export function AppSidebar() {
...
-      } h-screen p-4 shadow-lg border-r-2 flex flex-col justify-between transition-all duration-300 items-center relative`}
+      className={`bg-white ${isExpanded ? "w-52 items-start" : "w-18 items-center"
+        } h-screen p-4 shadow-lg border-r-2 flex flex-col justify-between transition-all duration-300 items-center relative`}
     >
       <div
-        className={`flex ${
-          isExpanded ? "items-start" : "items-center"
-        }  flex-col`}
+        className={`flex ${isExpanded ? "items-start" : "items-center"
+          }  flex-col`}
```
arseny-kostenko: seems this fits on one line

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -11,7 +11,7 @@ import type { Store } from "@reduxjs/toolkit"
 export const connectionEpic = (store: Store) => action$.pipe(
     filter((action) => {
         const state = store.getState()
-        return !selectConnected(state) && action.type === setConnecting.type
+        return selectStatus(state) != "Connected" && action.type === connectPending.type
```
arseny-kostenko: "Connected" should become a constant

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -32,10 +32,21 @@ export const sendRequestEpic = () =>
...
             const socket = getSocket()
             socket.next({ type: setData.type, payload: undefined })
         }),
     )
 
+export const disconnectEpic = () =>
+    action$.pipe(
+        filter((action) => action.type === resetConnection.type),
+        tap((action) => {
+            console.log("We're inside the disconnect epic")
```
arseny-kostenko: do you need this?

apps/frontend/src/state/epics/wsConnectionEpic.ts
```diff
@@ -50,14 +50,15 @@ export const wsConnectionEpic = (store: Store) =>
                 ignoreElements(),
                 takeUntil(
                     action$.pipe(
-                        filter((action) => action.type === 'wsconnection/disconnect'), // Need to fix this
+                        filter((action) => action.type === VALKEY.CONNECTION.closeConnection),
                         tap(() => {
+                            console.log('Socket Connection closed')
```
arseny-kostenko: need this?

apps/frontend/src/state/epics/wsConnectionEpic.ts
```diff
@@ -9,18 +9,19 @@ import {
...
+    connectRejected,
 } from '@/state/wsconnection/wsConnectionSlice'
 import { action$ } from '../middleware/rxjsMiddleware/rxjsMiddlware'
 import type { PayloadAction, Store } from '@reduxjs/toolkit'
+import { VALKEY } from "@common/src/constants.ts"
 
 let socket$: WebSocketSubject<PayloadAction> | null = null;
 
 export const wsConnectionEpic = (store: Store) =>
     action$.pipe(
```
arseny-kostenko: it looks to me that there are two (at least) epics in this epic — one that creates the socket and another that starts accepting messages from this socket.

I would split it correspondingly, and instead of one massive epic, create a root epic that merges those individual ones with their concerns the same way you do in `registerEpics`:
```
export const wsConnectionEpic = store => merge(
  createSocket,
  emitActions,
  retryConnections, // which you will need to add
)
```

so that instead of one giant mergeMap, you'll have some isolated mini-epics that are really easy to reason about

package-lock.json
```diff
@@ -12,9 +12,6 @@
         "apps/server",
         "common/*"
       ],
-      "dependencies": {
```
arseny-kostenko, line 15: it appears that you run `npm i` — you should be running `npm ci` — it will install all packages and their dependencies exactly as it's specified in package-lock.json, because otherwise, npm i will recalculate it from scratch and can change the dependency tree — it's what we're observing here. Dependency tree should be stable, so we can reproduce issues, fix vulnerabilities using `npm audit fix` and other deterministic reasons. More details here https://support.deploybot.com/build-tools/why-developers-should-use-npm-ci-instead-of-npm-install-and-its-benefits

PR #16, opened 2025-09-10
fixed ui related comments

nassery318:


apps/frontend/src/components/ui/app-sidebar.tsx
```diff
@@ -69,7 +69,7 @@ export function AppSidebar() {
               <li key={to}>
                 <Link
                   to={to}
-                  className={`flex p-2 text-nowrap dark:text-white ${getNavItemClasses(to)} h-10`}
+                  className={`flex p-2 text-nowrap dark:text-white ${getNavItemClasses(to)} h-10 ${isExpanded && "w-42"}`}
```
arseny-kostenko: no no, you don't need to hardcode width of these elements. All you need is to make their container stretch (in both places)
Screenshot 2025-09-10 at 9 39 39 AM

apps/frontend/src/components/ui/app-sidebar.tsx
```diff
@@ -69,11 +67,15 @@ export function AppSidebar() {
               <li key={to}>
                 <Link
                   to={to}
-                  className={`flex p-2 text-nowrap dark:text-white ${getNavItemClasses(to)} h-10 ${isExpanded && "w-42"}`}
+                  className={`flex p-2 dark:text-white ${getNavItemClasses(
+                    to
```
arseny-kostenko, line 71: your prettify is wild lol

apps/frontend/src/components/ui/app-sidebar.tsx
nassery318: made the fix. stretch was causing a bitter of jitters when expanding. Now it is fixed by bringing in the menu item and icon inside a div rather than a span.

apps/frontend/src/components/ui/app-sidebar.tsx
nassery318: :)

PR #17, opened 2025-09-12
github workflow to add lint rules

ali-sadri:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/checkout@v4
+
+      # Step 2: Set up Node.js environment
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20" # Specify the Node.js version
+
+      # Step 3: Install dependencies
+      - name: Install dependencies
```
nassery318, line 41: Maybe we add a node modules caching step before the step 3, it will speed up the build

.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: npm install
```
nassery318: Lets do "npm ci" rather than "npm install" and rely on the existing package-lock.json to keep libraries consistent.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,35 @@
...
+      # Step 1: Check out the code
+      # This checks out the code from the pull request branch
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Step 2: Set up Node.js environment
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20" # Specify the Node.js version
```
arseny-kostenko: node 20 is getting obsolete in a few months

ali-sadri: Github recommends the node version to match the node.js used in our project locally. Aren't we using version 20? Which version do you recommend to be used here.

arseny-kostenko: can't speak for everyone but I use 22 

https://nodejs.org/en/about/previous-releases

.github/workflows/lint.yml
ali-sadri: pushed the change to PR.

.github/workflows/lint.yml
ali-sadri: pushed the change to PR

.github/workflows/lint.yml
ali-sadri: pushed the change to PR

PR #19, opened 2025-09-16
Render dashboard based on connectionId

ravjotbrar:


apps/frontend/src/state/valkey-features/info/infoSelectors.ts
```diff
@@ -1,4 +1,7 @@
 import type { RootState } from "@/store.ts";
-import {VALKEY} from "@common/src/constants.ts"
+import { VALKEY } from "@common/src/constants.ts"
+
+export const selectData = (id: string) => (state: RootState) =>
+    state[VALKEY.STATS.name]?.[id]?.data ?? null
```
arseny-kostenko: you don't need to change it but that's why I used path from ramda in another place like this
```
R.path([VALKEY.COMMAND.name, id, "commands", index]),
```

apps/frontend/src/state/valkey-features/info/infoSlice.ts
```diff
@@ -1,39 +1,76 @@
...
-            state.data.bytes_per_key = action.payload.memory['keys.bytes-per-key'];
-            state.data.server_name = action.payload.info['server_name'];
-            state.data.tcp_port = action.payload.info['tcp_port'];
+            const { connectionId, info, memory } = action.payload
+            if (!state[connectionId]) {
+                state[connectionId] = createInitialConnectionState()
+            }
+
+            state[connectionId].data.total_commands_processed = info["total_commands_processed"]
+            state[connectionId].data.connected_clients = info['connected_clients']
```
arseny-kostenko: double quotes from here to below

arseny-kostenko: but nevermind, how about we do
```
state[connectionId].data = R.applySpec({
  dataset_bytes: R.path(["memory", "dataset.bytes"]),
  keys_count: R.path(["memory", "keys.count"]),
  bytes_per_key: R.path(["memory", "keys.bytes-per-key"]),
  server_name: R.path(["info", "server_name"]),
  tcp_port: R.path(["info", "tcp_port"]),
  total_commands_processed: R.path(["info", "total_commands_processed"]),
  connected_clients: R.path(["info", "connected_clients"])
})(action.payload)
```

PR #21, opened 2025-09-18
implementing initial end to end stages of key browser

nassery318:


apps/frontend/src/state/valkey-features/keys/keyBrowserSelectors.ts
```diff
@@ -0,0 +1,23 @@
+import type { RootState } from "@/store.ts"
+import * as R from "ramda"
+import {VALKEY} from "@common/src/constants.ts"
+
+export const selectKeys = (id: string) => (state: RootState) =>
+  R.path([VALKEY.KEYS.name, id, "keys"], state) || []
```
arseny-kostenko: you can do [pathOr](https://ramdajs.com/docs/#pathOr)
```
R.pathOr([], [VALKEY.KEYS.name, id, "keys"], state)
```

arseny-kostenko: and for the rest of them below

apps/frontend/src/state/valkey-features/keys/keyBrowserSlice.ts
```diff
@@ -0,0 +1,67 @@
...
+const initialState: KeyBrowserState = {}
+
+const keyBrowserSlice = createSlice({
+  name: "keyBrowser",
+  initialState,
+  reducers: {
+    getKeysRequested: (state, action: PayloadAction<{ connectionId: string; pattern?: string; count?: number }>) => {
+      const { connectionId } = action.payload
+      if (!state[connectionId]) {
+        state[connectionId] = {
```
arseny-kostenko: I see you have this default empty state above as well — let's make it `export const` and reuse in both places

apps/frontend/src/main.tsx
```diff
@@ -12,6 +12,7 @@ import RequireConnection from "./components/RequireConnection.tsx"
 import Settings from "./components/Settings.tsx"
 import LearnMore from "./components/LearnMore.tsx"
 import "./css/index.css"
+import { KeyBrowser } from "./components/KeyBrowser.tsx"
```
arseny-kostenko, line 14: could you please move this above the css, thanks

apps/frontend/src/components/KeyBrowser.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <div className="p-4">
+      <AppHeader title="Key Browser" icon={<Compass size={20} />} />
+
+      {loading && <div className="ml-2">Loading keys...</div>}
+      {error && <div className="ml-2">Error loading keys: {error}</div>}
+
+      {/* Total Keys and Key Stats */}
+      <div className="flex justify-between">
+        <div className="h-20 w-1/4 p-4 dark:border-tw-dark-border border-1 rounded flex flex-col justify-center items-center">
```
ravjotbrar: In the documentation, I see that tailwind has a border and border-2 but no border-1. Is that true? https://tailwindcss.com/docs/border-width#basic-example

apps/frontend/src/components/KeyBrowser.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { useAppDispatch } from "@/hooks/hooks";
+import { useParams } from "react-router";
+import { AppHeader } from "./ui/app-header";
+import { Compass, RefreshCcw } from "lucide-react";
+
+export function KeyBrowser() {
+  const { id } = useParams();
+  const dispatch = useAppDispatch();
+
+  const keys = useSelector(selectKeys(id || ""));
```
ravjotbrar: Since id should always be defined when selecting keys, should we add a check for it here instead of defaulting to empty string?

arseny-kostenko: ah, you're right, it should be
```
const keys = useSelector(selectKeys(id!))
```
to satisfy the ts compiler

apps/frontend/src/state/valkey-features/keys/keyBrowserSlice.ts
nassery318: created the defaultConnectionState and used it across the slice and selector

apps/frontend/src/state/valkey-features/keys/keyBrowserSelectors.ts
nassery318: used R.pathOr

apps/frontend/src/main.tsx
nassery318: Moved on top of CSS

apps/frontend/src/components/KeyBrowser.tsx
nassery318: You re right there is no border-1 there is border which adds a 1px border

apps/frontend/src/components/KeyBrowser.tsx
nassery318: resolved with id!

PR #24, opened 2025-09-23
extending the key information and improving the key browser display

nassery318:


apps/frontend/src/components/KeyBrowser.tsx
```diff
@@ -46,12 +48,20 @@ export function KeyBrowser() {
   const handleKeyClick = (keyName: string) => {
     setSelectedKey(keyName);
 
-    const keyInfo = keys.find((k) => k.key === keyName);
+    const keyInfo = keys.find((k) => k.name === keyName);
     if (keyInfo && !keyInfo.type) {
```
arseny-kostenko: I know you added it in the previous PR, but it's a bit confusing — is this basically a check that keyInfo actually has data? and for some reason find can return an empty object instead of an actual keyInfo? if so, it's semantically more straightforward for ~future generations~ the reader to see if (R.isNotEmpty(keyInfo)) https://ramdajs.com/docs/#isNotEmpty

apps/server/src/keys-browser.ts
```diff
@@ -2,6 +2,66 @@ import {WebSocket} from "ws"
...
+
+    const keyInfo: EnrichedKeyInfo = {
+      name: key,
+      type: keyType,
+      ttl: ttl,
+      size: memoryUsage || 0
+    }
+
+    // Get collection size per type
+    try {
```
arseny-kostenko, line 30: not that important, but it I think it can be less verbose if you use a Record:
```
try {
  const sizeCommands: Record<string, string> = {
    list: "LLEN",
    set: "SCARD",
    zset: "ZCARD",
    hash: "HLEN",
    stream: "XLEN"
  }

  const command = sizeCommands[keyType?.toLowerCase() ?? ""]
  if (command) {
    keyInfo.collectionSize = await client.customCommand([command, key]) as number
  }
  // else string (or unknown) has no collection size
} catch (err) {
  console.log(`Could not get collection size for key ${key}:`, err)
}
```

arseny-kostenko: also, can a keyType really be null | undefined for these types? it seems like unnecessarily defensive programming. let it fail here in the catch block because otherwise — it propagates to UI

apps/server/src/keys-browser.ts
```diff
@@ -2,6 +2,66 @@ import {WebSocket} from "ws"
...
+      }
+    } catch (err) {
+      console.log(`Could not get collection size for key ${key}:`, err)
+    }
+
+    return keyInfo
+  } catch (err) {
+    return {
+      name: key,
+      type: 'unknown',
```
arseny-kostenko: let's stay consistent if quotes — it's double quotes we agreed on?

apps/server/src/keys-browser.ts
```diff
@@ -23,14 +83,38 @@ export async function getKeys(client: GlideClient, ws: WebSocket, payload: {
       
       console.log("SCAN response:", rawResponse)
       
-      // SCAN Results [cursor, [keys...]]
       const [cursor, keys] = rawResponse
       
+      const enrichedKeys: EnrichedKeyInfo[] = []
+      
+      if (keys && keys.length > 0) {
```
arseny-kostenko: this is an unnecessary check if you use monadic types like `Array` instead of `for` loops, and full null safety you can either use `?` or default values `const [cursor, keys = []] = rawResponse`, so then you can rewrite it more declaratively and concisely:

```
const batchSize = 10 // todo make it configurable

const enrichedKeys = R.flatten(
  await Promise.all(
    R.splitEvery(batchSize, keys).map(async batch => {
      const settled = await Promise.allSettled(batch.map(k => getKeyInfo(client, k)))
      return settled.map((res, idx) =>  res.status === 'fulfilled'
          ? res.value
          : { name: batch[idx], type: 'unknown', ttl: -1, size: 0 })
    })
  )
)
```

common/src/bytes-conversion.ts
```diff
@@ -0,0 +1,7 @@
+export function formatBytes (bytes: number) {
+    if (bytes === 0) return '0 B';
```
arseny-kostenko: tell chatgpt to always use semicolons and string templates instead of string concatenation ;)

common/src/memory-usage-calculation.ts
```diff
@@ -0,0 +1,11 @@
+interface KeyInfo {
+    name: string;
+    type: string;
+    ttl: number;
+    size: number;
+    collectionSize?: number;
+  }
+
+export function calculateTotalMemoryUsage (keys: KeyInfo[]) {
```
arseny-kostenko, line 9: the file name in nodejs should be kebab case

nassery318: memory-usage-calculation.ts changed to kebab case

common/src/bytes-conversion.ts
nassery318: will do 😀

apps/server/src/keys-browser.ts
nassery318: changed to `const command = sizeCommands[keyType.toLowerCase()]` and letting it fail rather than being defensive as key always has a TYPE.

apps/server/src/keys-browser.ts
nassery318: "" everywhere

apps/frontend/src/components/KeyBrowser.tsx
nassery318: The keyInfo && check ensures that find() found a match - Array.find() returns undefined when no match is found (never an empty object). The !keyInfo.type part is an to avoid unnecessary API calls when we already have the type data. True it will be more readable (R.isNotEmpty(keyInfo) && !keyInfo!.type) and will work, but with some extra type gymnastics.

apps/server/src/keys-browser.ts
nassery318: resolved with more concise version suggested

PR #26, opened 2025-09-25
implemented key deletion with UNLINK and custom delete modal

nassery318:


apps/server/src/index.ts
```diff
@@ -86,6 +86,24 @@ wss.on("connection", (ws: WebSocket) => {
           })
         );
       }
+    } else if (action.type === VALKEY.KEYS.deleteKeyRequested) {
+      console.log("Handling deleteKeyRequested for key:", action.payload?.key);
```
ravjotbrar, line 90: I'm not sure if this is a github formatting issue, but it looks like the conditional blocks are not indented.

ravjotbrar: Ignore this. It was a ravjot issue.

nassery318: It is a good observation, and points out that we need to have formatting rules.

PR #29, opened 2025-09-26
Add linting pipeline and fix all linting errors in project

ravjotbrar:


eslint.config.js
```diff
@@ -28,6 +28,7 @@ export default tseslint.config([
       globals: globals.browser,
     },
     rules: {
+      "@typescript-eslint/no-explicit-any": "off",
```
ArgusLi: Are we no longer checking for explicit anys?

eslint.config.ts
ArgusLi, line 1: Why was this file deleted?

ravjotbrar: All the rules were defined in eslint.config.js, so I kept that one. Having two is redundant, but I can rename our current one to a ts file.

eslint.config.js
ravjotbrar: I really didn't want to spend too much time try to fix the explicit anys in our project, so I turned it off for now. But we could ignore locally in the spots that need fixing and leave them as TODOs if that helps.

ArgusLi: Sounds good!

eslint.config.ts
ArgusLi: Ah, if it's redundant then there's no need to keep it. Thanks!

PR #32, opened 2025-10-01
Delete connection logic

ravjotbrar:


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -48,6 +48,30 @@ export const connectionEpic = (store: Store) =>
     )
   )
 
+export const deleteConnectionEpic = () => 
+  action$.pipe(
+    select(deleteConnection),
+    tap(({ payload: { connectionId } }) => {
+      try {
+        const currentConnections = R.pipe(
```
arseny-kostenko, line 57: is it time to extract this into a separate function? there is at least one more place where we do that. can do later though but ideally having a // todo

apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
```diff
@@ -74,8 +74,22 @@ const connectionSlice = createSlice({
         ...action.payload,
       }
     },
+    deleteConnection: (state, action) => {
+      const { connectionId } = action.payload
+      if(state.connections[connectionId]) {
```
arseny-kostenko: can do just return `R.dissocPath(["connections", connectionId], state)`

and there is extra newline below

PR #33, opened 2025-10-01
Add new keys - String & Hash

nassery318:
Keys adding functionality added so far:

1. String
2. Hash

### DIFF
--- a/common/src/key-validators.ts
+++ b/common/src/key-validators.ts
@@ -1,63 +1,63 @@
 interface ValidationData {
-    keyName: string
-    keyType: string
-    value?: string
-    ttl?: number
-    hashFields?: Array<{ field: string; value: string }>
+  keyName: string
+  keyType: string
+  value?: string
+  ttl?: number
+  hashFields?: Array<{ field: string; value: string }>
 }
 
 interface ValidationRule {
-    validatorFn: (key: ValidationData) => boolean
-    error: string
+  validatorFn: (key: ValidationData) => boolean
+  error: string
 }
 
 const isBlank = (value: string | undefined | null): boolean =>
-    !value || value.trim().length === 0
+  !value || value.trim().length === 0
 
 const isNotBlank = (value: string | undefined | null): boolean =>
-    !isBlank(value)
+  !isBlank(value)
 
 const TtlSpec: ValidationRule[] = [
-    {
-        validatorFn: (key) => isNotBlank(key.ttl?.toString()) && (isNaN(key.ttl!) || key.ttl! < -1),
-        error: "TTL is not a valid number (-1 for no expiration, or a positive number)"
-    },
+  {
+    validatorFn: (key) => isNotBlank(key.ttl?.toString()) && (isNaN(key.ttl!) || key.ttl! < -1),
+    error: "TTL is not a valid number (-1 for no expiration, or a positive number)",
+  },
 ]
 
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/ui/add-key.tsx
```diff
@@ -0,0 +1,264 @@
...
+    index: number,
+    key: "field" | "value",
+    val: string
+  ) => {
+    const updated = [...hashFields]
+    updated[index][key] = val
+    setHashFields(updated)
+  }
+
+  const handleSubmit = (e: React.FormEvent) => {
```
arseny-kostenko, line 43: this function here is getting a bit out of hand due to having the validation logic here, and this are just string and hash types. I'd suggest creating a module in common/src like key-validators.ts or something like this and move all validators per type there in a declarative way, something along the lines of:
```
const TtlSpec = [
  {
    validatorFn: key => isNotBlank(key.ttl!) && (isNaN(key.ttl!) || key.ttl! < -1),
    error: "TTL is not a valid number (-1 for no expiration, or a positive number)"
  },
]

const BaseSpec = [
   ...TtlSpec,
  { validatorFn: key => isBlank(key.keyName?.trim()), error: "Key name is required" },
  { validatorFn: key => isBlank(key.value), error: "Value is required" },
]

const StringSpec = [ 
  ...BaseSpec,
  { validatorFn, error },
  { validatorFn, error },
  ...
]
 
const HashSpec = [ 
  ...BaseSpec,
  { validatorFn, error },
  { validatorFn, error },
  ...
]

// spec => key => error:string
const validate = spec => key => spec
  .filter(({ validatorFn, error }) => validatorFn(key))
  .reduce((acc, { error }) => acc.concat(`${error}; `), "")

const validators = {
   "string": validate(StringSpec),
   "hash": validate(HashSpec),
    "undefined":  () => { error: "Key type is required" },
   ... 
}
```
so that in your `handleSubmit` you can be just
```
const errors = validators[keyType]({ keyName, keyType, value, ttl: parseTtl })
if (R.isNotEmpty(errors) return setError(errors)
```

we'll probably need this logic on the backend anyway to validate there too

arseny-kostenko: can make validate a single reduce fn too

apps/frontend/src/components/ui/add-key.tsx
```diff
@@ -0,0 +1,264 @@
...
+          <div className="pt-2 text-sm flex gap-4">
+            <button
+              className="px-4 py-2 w-full bg-tw-primary text-white rounded hover:bg-tw-primary/90"
+              disabled={keyType === "Select key type" || !keyName}
+              type="submit"
+            >
+              Submit
+            </button>
+            <button
+              className="px-4 py-2 w-full bg-tw-primary text-white rounded hover:bg-tw-primary/90"
```
arseny-kostenko: maybe let's have Cancel as a regular, not primary, button, so that we have only one CTA?

nassery318: image

apps/frontend/src/components/ui/add-key.tsx
nassery318: True, since adding more types will increase the size of handleSubmit even more as suggested I created the key-validators.ts function and moved validation there.

PR #35, opened 2025-10-04
Add clusterSlice and cluster compatibility

ravjotbrar:


apps/server/src/index.ts
```diff
@@ -144,6 +144,20 @@ async function connectToValkey(
 
     clients.set(payload.connectionId, client)
 
+    if(await belongsToCluster(client)) {
+      const clusterNodes = await discoverCluster(client)
+      if (!clusterNodes || Object.keys(clusterNodes).length === 0) {
```
arseny-kostenko: it looks like discoverCluster always returns an object, so you don't really need `!clusterNodes` and can just do `isNotEmpty(clusterNodes)`

apps/server/src/index.ts
```diff
@@ -168,6 +182,55 @@ async function connectToValkey(
   }
 }
 
+async function belongsToCluster(client: GlideClient): Promise<boolean> {
+  const response = await client.info([InfoOptions.Cluster])
+  const parsed = parseInfo(response)
+  return parsed["cluster_enabled"] === "1" ?  true : false
```
arseny-kostenko: unnecessary ternary, can just `return parsed["cluster_enabled"]`

apps/server/src/index.ts
ravjotbrar: Without it, typescript complains

apps/server/src/index.ts
ravjotbrar: Wouldn't "0" be truthy?

arseny-kostenko: no, because you're using strict equality `===`, not `==`

apps/server/src/index.ts
arseny-kostenko: even with exclamation mark? `isNotEmpty(clusterNodes!)`

ravjotbrar: That worked

apps/server/src/index.ts
ravjotbrar: If I `return parsed["cluster_enabled"]` as you're suggesting, that will return either "1" or "0". 

Both return values will always be true in an if block, unless I'm missing something?

arseny-kostenko: my bad, I meant you can `return parsed["cluster_enabled"] === "1"` — without a ternary

ravjotbrar: Ahh good catch I'll change it

apps/frontend/src/components/connection/ConnectionEntry.tsx
```diff
@@ -60,14 +61,14 @@ export const ConnectionEntry = ({ connectionId, connection }: ConnectionEntryPro
...
-          <Link title={label} to={`/${connectionId}/dashboard`}>{label}</Link>
+          <Link title={label} to={(clusterId ? `/${clusterId}/${connectionId}/dashboard` : `/${connectionId}/dashboard`)}>{label}</Link>
         </Button>
       }
       <div className="flex flex-row justify-end">
         {
           isConnected &&
           <>
-            <Button onClick={() => history.navigate(`/${connectionId}/dashboard`)} variant="ghost">
+            <Button onClick={() => history.navigate(clusterId ? `/${clusterId}/${connectionId}/clustertopology` : `/${connectionId}/dashboard`)} variant="ghost">
```
arseny-kostenko: `cluster-topology`

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -84,8 +84,11 @@ export const sendRequestEpic = () =>
 export const setDataEpic = () =>
   action$.pipe(
     select(connectFulfilled),
-    tap(({ payload: { connectionId } }) => {
+    tap((action) => {
       const socket = getSocket()
-      socket.next({ type: setData.type, payload: { connectionId } })
+      const { clusterId, connectionId } = action.payload
+      if(clusterId)
```
arseny-kostenko: missing space

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -84,8 +84,11 @@ export const sendRequestEpic = () =>
...
+        socket.next({ type: setClusterData.type, payload: { clusterId, connectionId } })
+      else socket.next({ type: setData.type, payload: { connectionId } })
```
arseny-kostenko, line 91: let's keep the layout consistent, do either
```
if socket.next
else socket.next
```

or
```
if ()
  socket.next
else
  socket.next
```

or even 
```
const nextAction = clusterId ? { type, payload: { clusterId, connectionId } } : { type, payload: { connectionId } }
socket.next(nextAction)
```

apps/server/src/index.ts
```diff
@@ -48,19 +47,29 @@ wss.on("connection", (ws: WebSocket) => {
     }
     if (action.type === VALKEY.STATS.setData) {
       const client = clients.get(connectionId)
-      if (client) {
+
+      if(client instanceof GlideClient)
```
arseny-kostenko: missing space. and below. do you have some setting in IDE that does this?

apps/server/src/index.ts
```diff
@@ -171,23 +182,37 @@ async function connectToValkey(
...
-    const client = await GlideClient.createClient({
+    const standaloneClient = await GlideClient.createClient({
       addresses,
       requestTimeout: 5000,
       clientName: "test_client",
     })
 
-    clients.set(payload.connectionId, client)
-
+    if(await belongsToCluster(standaloneClient)) {
```
arseny-kostenko: space

apps/server/src/index.ts
```diff
@@ -212,13 +237,99 @@ async function connectToValkey(
...
+}
+
+async function discoverCluster(client: GlideClient) {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const response = await client.customCommand(["CLUSTER", "SLOTS"]) as any[][]
+
+    const firstSlotRange = response[0]
+    const firstPrimaryNode = firstSlotRange?.[2]
+    const clusterId = firstPrimaryNode?.[2]
```
arseny-kostenko: this is an implicit nesting, i.e. to understand where the clusterId comes from, a future reader will need to understand where firstPrimiaryNode comes from, and then where the slot comes from, so better rewrite it explicitly and add comments:
```
const { firstSlotRange, firstPrimaryNode, clusterId } = R.applySpec({
  firstSlotRange: R.nth(0),
  firstPrimaryNode: R.path([0, 2]), // firstSlotRange[2]
  clusterId: R.path([0, 2, 2]), // firstPrimaryNode[2]
})(response)
```

apps/server/src/index.ts
```diff
@@ -212,13 +237,99 @@ async function connectToValkey(
...
+
+    const clusterNodes = response.reduce((acc, slotRange) => {
+      const [, , ...nodes] = slotRange
+
+      nodes.forEach((node, index) => {
```
arseny-kostenko: forEach exists for side effects, it's signature is t -> void, just like `tap` operator
here, you're not doing a nested reduce in fact. so it's something like
```
const clusterNodes = response.reduce((acc, slotRange) => {
      const [, , ...nodes] = slotRange
      
      return nodes.reduce((nodesById, [host, port], idx) => {
         const id = `${host}-${port}`
         return nodesById[id]
               ? nodesById
               :  { ...nodesById, [id]: { host, port, role: idx === 0 ? "primary" : "replica" } }
  }, acc)
```

apps/server/src/index.ts
```diff
@@ -212,13 +237,99 @@ async function connectToValkey(
...
+
+async function connectToCluster(
+  standaloneClient: GlideClient, 
+  ws: WebSocket, 
+  clients: Map<string, GlideClient | GlideClusterClient>, 
+  payload: { host: string; port: number; connectionId: string;},
+  addresses: { host: string, port: number | undefined }[],
+) {
+  const { clusterNodes, clusterId } = await discoverCluster(standaloneClient)
+  if (Object.keys(clusterNodes!).length === 0) {
```
arseny-kostenko: R.isEmpty?

apps/server/src/index.ts
```diff
@@ -243,6 +354,63 @@ async function setDashboardData(
...
+
+type ParsedClusterInfo = {
+  [host: string]: {
+    [section: string]: {
+      [key: string]: string
+    }
+  }
+}
+
+function parseClusterInfo(rawInfo: ClusterResponse<string>): ParsedClusterInfo {
```
arseny-kostenko: this is a very imperative code — I had to read it several times to understand. But it ultimately, it's a data transformation pipeline:
```
rawInfo -> [host, infoString] -> [host, infoObj] -> set<host, infoObj>
```
so I would suggest to write it like a `pipe`line:

```
export const parseClusterInfo = (rawInfo: ClusterResponse<string>): ParsedClusterInfo =>
  R.pipe(
    R.toPairs, // [ [host, infoString], ... ]
    R.map(([host, infoString]) =>
      [
        host,
        R.pipe(
          R.split("\r\n"),
          R.reduce(
            (
              state: { currentSection: string | null; hostData: ParsedClusterInfo[string] },
              line: string
            ) => {
              const trimmed = line.trim()
              if (trimmed === "") return state

              if (trimmed.startsWith("# ")) {
                const section = trimmed.slice(2).trim()
                return { currentSection: section, hostData: { ...state.hostData, [section]: {} } }
              }

              if (!state.currentSection) return state

              const [key, ...rest] = line.split(":")
              const value = rest.join(":")
              return {
                currentSection: state.currentSection,
                hostData: {
                  ...state.hostData,
                  [state.currentSection]: {
                    ...(state.hostData[state.currentSection] ?? {}),
                    [key]: value
                  },
                },
              },
            },
            { currentSection: null as string | null, hostData: {} as ParsedClusterInfo[string] }
          ),
          (s: { hostData: ParsedClusterInfo[string] }) => s.hostData
        )(infoString)
      ] as [string, ParsedClusterInfo[string]]
    ),
    R.fromPairs,
  )(rawInfo)
  ```

File: apps/server/src/index.ts (hunk 6 above)
Comment (ravjotbrar) [reply apps/server/src/index.ts]: I'll add a rule to our eslint config

File: apps/server/src/index.ts (hunk 11 above)
Comment (ravjotbrar) [reply apps/server/src/index.ts]: This might be a case where we agree to disagree, because I have to read the functional code multiple times to understand it. Also isn't the above code less efficient because of the spread syntax creating object copies?

I'll let you make the executive decision here, since this isn't a hill I'm willing to die on.

Comment (arseny-kostenko) [reply apps/server/src/index.ts]: you're right about destructuring, it can be fixed though
```
export const parseClusterInfo = (rawInfo: ClusterResponse<string>): ParsedClusterInfo =>
  R.pipe(
    R.toPairs, // [ [host, infoString], ... ]
    R.map(([host, infoString]) =>
      [
        host,
        R.pipe(
          R.split("\r\n"),
          R.reduce(
            (
              state: { currentSection: string | null; hostData: ParsedClusterInfo[string] },
              line: string
            ) => {
              const trimmed = line.trim()
              if (trimmed === "") return state

            if (trimmed.startsWith("# ")) {
              const section = trimmed.slice(2).trim()
              state.currentSection = section
              state.hostData[section] = state.hostData[section] || {}
              return state
            }
          
            if (!state.currentSection) return state
          
            // split only once (values may contain ':')
            const idx = line.indexOf(":")
            if (idx === -1) return state
            const key = line.slice(0, idx)
            const value = line.slice(idx + 1)
          
            state.hostData[state.currentSection] = state.hostData[state.currentSection] || {}
            state.hostData[state.currentSection]![key] = value
            return state
            { currentSection: null as string | null, hostData: {} as ParsedClusterInfo[string] }
          ),
          (s: { hostData: ParsedClusterInfo[string] }) => s.hostData
        )(infoString)
      ] as [string, ParsedClusterInfo[string]]
    ),
    R.fromPairs,
  )(rawInfo)
  ```

arseny-kostenko: the point is not functional vs imperative, the point is that to understand _**that specific**_ imperative code — you cannot read it from top to bottom once — you gotta scan it down and then up, and then down again

ravjotbrar: Changed

PR #38, opened 2025-10-09
Update set and list key types

nassery318:


apps/server/src/keys-browser.ts
```diff
@@ -391,6 +391,39 @@ async function updateHashKey(
...
 
+async function updateListKey(
+  client: GlideClient,
+  key: string,
+  updates: { index: number; value: string }[],
+  ttl?: number
+) {
+  // Update each specified index
+  for (const { index, value } of updates) {
+    await client.customCommand(["LSET", key, index.toString(), value])
```
arseny-kostenko: can you try making it a bulk update instead of a sequence of multiple calls? like this:
```
const pipeline = client.pipeline()

for (const { index, value } of updates) {
  pipeline.customCommand(["LSET", key, index.toString(), value])
}

await pipeline.exec()
```

apps/server/src/keys-browser.ts
```diff
@@ -391,6 +391,39 @@ async function updateHashKey(
...
+}
+
+async function updateSetKey(
+  client: GlideClient,
+  key: string,
+  updates: { oldValue: string; newValue: string }[],
+  ttl?: number
+) {
+  // For each update, remove old value and add new value
+  for (const { oldValue, newValue } of updates) {
```
arseny-kostenko, line 421: same here

apps/server/src/keys-browser.ts
nassery318: https://github.com/valkey-io/valkey-glide/wiki/NodeJS-wrapper

By using batch (Atomic Batch) to execute commands in one transaction. 

```
const batch = new Batch(true)

  updates.map(({ index, value }) =>
    batch.customCommand(["LSET", key, index.toString(), value])
  )
```

apps/server/src/keys-browser.ts
nassery318: same here using Atomic Batch

PR #42, opened 2025-10-21
Build dashboard UI and add metrics + Search and Filtering

nassery318:


common/src/format-metric-value.ts
```diff
@@ -0,0 +1,45 @@
+import { formatBytes } from "./bytes-conversion"
+
+export type ValueType = "bytes" | "number" | "mixed"
+
+const formatSeconds = (seconds: number): string => {
```
arseny-kostenko: feels like this belongs in common/src/time-utils.ts

apps/frontend/src/components/Dashboard.tsx
```diff
@@ -1,40 +1,165 @@
...
 
 export function Dashboard() {
   const { id, clusterId } = useParams()
-  const instanceData = useSelector(selectData(id!))
+  const infoData = useSelector(selectData(id!))
   const clusterData = useSelector(selectClusterData(clusterId!))
   const connectionDetails = useSelector(selectConnectionDetails(id!))
   const nodeAddress = `${connectionDetails.host}:${connectionDetails.port}`
   console.log("The cluster data is: ", clusterData)
   console.log("The node address is: ", nodeAddress)
```
ravjotbrar: These could both be removed, I forgot them during my PR

apps/frontend/src/components/Dashboard.tsx
```diff
@@ -1,40 +1,165 @@
...
 import { selectClusterData } from "@/state/valkey-features/cluster/clusterSelectors"
 import { selectConnectionDetails } from "@/state/valkey-features/connection/connectionSelectors"
 
 export function Dashboard() {
   const { id, clusterId } = useParams()
-  const instanceData = useSelector(selectData(id!))
+  const infoData = useSelector(selectData(id!))
   const clusterData = useSelector(selectClusterData(clusterId!))
   const connectionDetails = useSelector(selectConnectionDetails(id!))
   const nodeAddress = `${connectionDetails.host}:${connectionDetails.port}`
```
ravjotbrar: Could probably remove these as well? Don't think they're used.

apps/frontend/src/components/Dashboard.tsx
nassery318: Removed

apps/frontend/src/components/Dashboard.tsx
nassery318: Yes thees are not used. Removed for now

PR #45, opened 2025-10-23
build a donut chart for key types and memory usage per each key type

nassery318:


apps/frontend/src/components/ui/donut-chart.tsx
```diff
@@ -0,0 +1,131 @@
...
+    name: type.charAt(0).toUpperCase() + type.slice(1),
+    count: stats.count,
+    totalSize: stats.totalSize,
+    percentage: ((stats.count / keys.length) * 100).toFixed(1),
+  }))
+
+  const totalKeys = keys.length
+  const totalMemory = calculateTotalMemoryUsage(keys)
+
+  const chartColors: Record<string, string> = {
```
arseny-kostenko, line 67: let's stick to the existing tailwind's palette for consistency , something like this:

```
import colors from "tailwindcss/colors"

export const chartColors = {
  Set: colors.teal[600],
  Hash: colors.indigo[500],
  List: colors.pink[500],
  String: colors.amber[500],
} satisfies Record<string, string>
```

PR #50, opened 2025-10-23
Create script to deploy cluster and update instructions

ArgusLi:


README.md
```diff
@@ -4,15 +4,10 @@
...
-
-    To start up and populate fake data: `docker compose --profile populate up --build`.
-
-    To start up without running the populate service:`docker compose up --build`.
-
-4. Add key value pairs using Redis Insight (Optional)
+1. From the root directory, run `tools/valkey-cluster/scripts/build_run_cluster.sh`.
+    1. This will build and run the cluster.
+    2. Master hostname will be `localhost`.
+    3. Master nodes ports will be `7001` - `7003`.
```
ravjotbrar, line 10: Is this always for certain?

ArgusLi: When looking at the cluster creation, it starts with 7001 then selects 2 more in addition as master. So it is guaranteed.

ArgusLi: `cluster_init.sh` line 77.

PR #54, opened 2025-10-28
Implement reconnect and retry when web socket is disconnected

nassery318:


apps/frontend/src/hooks/useWebSocketNavigation.ts
```diff
@@ -0,0 +1,37 @@
...
+    const wasConnected = previousStatus.current === CONNECTED
+    const isNowDisconnected = currentStatus === CONNECTING && wsConnection.reconnect.isRetrying
+    const isReconnecting = location.pathname === "/reconnect"
+
+    if (isReconnecting || location.pathname === "/connect") {
+      previousStatus.current = currentStatus
+      return
+    }
+
+    if (wasConnected && isNowDisconnected) {
```
arseny-kostenko, line 24: I would implement that in epic with chain/flatmap but since it's a single isolated hook, it's fine too

PR #56, opened 2025-10-31
Create monitor fetcher in valkey metrics server

ravjotbrar:


tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+        console.info("Hotkeys monitor connected")
+
+        const monitor = await monitorClient.monitor()
+        monitor.on("monitor", (_time, args) => processCommand(args))
+
+        // rest for a minute
+        await sleep(60000)
+
+        // compute top N keys
+        const sorted = [...keyCounts.entries()]
```
arseny-kostenko: fetcher should not do calculations, its only job is to get raw info and dump in a file. analytics should be ran separately against the files — when its endpoint is called

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
+import Valkey from "iovalkey"
+const url = "valkey://host.docker.internal:6379"
+//const url = String(process.env.VALKEY_URL || cfg.valkey.url || "").trim()
+export const makeHotkeysFetcher = (mainClient, topN = 10, prefixDelim = ":") => {
```
arseny-kostenko: don't do topN — we log everything in files, then run analytics against the files

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+      const prefix = parts[0]
+      prefixCounts.set(prefix, (prefixCounts.get(prefix) || 0) + 1)
+    }
+  }
+
+  const processCommand = args => {
+    if (!Array.isArray(args) || args.length < 2) return
+    console.log("Args are: ", args)
+    const command = String(args[0]).toLowerCase()
+    const key = args[1] ? String(args[1]) : null
```
arseny-kostenko: this doesn't account for `mget` that can retrieve multiple keys, we probably need separate parsers for different commands (as separate functions or within a single big functions — doesn't matter)

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+    if (key) recordKey(key)
+  }
+
+  const monitorLoop = async () => {
+    const monitorClient = new Valkey(url) 
+
+    while (true) {
+      try {
+        console.info("Hotkeys monitor connected")
```
arseny-kostenko: this probably should be after `await monitorClient.monitor()`

tools/valkey-metrics/config.yml
```diff
@@ -30,3 +30,8 @@ epics:
     type: "slowlog_len"
     poll_ms: 60000
     file_prefix: "slowlog_len"
+  
+  - name: "hotkeys"
+    type: "hotkeys"
+    poll_ms: 60000
```
arseny-kostenko: I presume, this is a duration of monitor? — I'd suggest renaming it because it's ambiguous.

We also need to configure interval between monitors. So like, monitoringDuration, monitoringInterval

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
+import Valkey from "iovalkey"
+const url = "valkey://host.docker.internal:6379"
```
arseny-kostenko: this probably shouldn't be here

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+          .sort((a, b) => b[1] - a[1])
+          .slice(0, topN)
+          .map(([key, count]) => ({ key, count, ts: Date.now() }))
+
+        lastSnapshot = sorted
+        console.info(`Captured top ${topN} hotkeys`)
+
+        await monitor.disconnect()
+        console.info("Hotkeys monitor sleeping for 1 minute")
+        await sleep(60_000)
```
arseny-kostenko: this should come from config.yml

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+      } catch (err) {
+        console.error("Hotkeys monitor error", err)
+        try { await monitorClient.disconnect() } catch {}
+        await sleep(5000)
```
arseny-kostenko: I would make this `Math.min(5000, 0.5 * interval)` so it's longer for long runs

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
```diff
@@ -0,0 +1,63 @@
...
+
+    while (true) {
+      try {
+        console.info("Hotkeys monitor connected")
+
+        const monitor = await monitorClient.monitor()
+        monitor.on("monitor", (_time, args) => processCommand(args))
+
+        // rest for a minute
+        await sleep(60000)
```
arseny-kostenko: this value should come from config.yml

tools/valkey-metrics/config.yml
arseny-kostenko: I would also add one more config property — maxCommandsPerRun: 1_000_000
because we can have idle nodes where nothing happens within the monitor duration but we can also have hot firehose nodes where you get millions of rows in 10 seconds — this will quickly lead to OOM. For statistical purposes, we'd be satisfied with 1M of rows within 5 seconds instead of waiting for 6 and get 30M or rows and a giant file

tools/valkey-metrics/src/effects/hotkeys-fetcher.js
arseny-kostenko: and like I mentioned above wrt maxCommandsPerRun — we need to constraint it this way as well, so I would add
```
let overflow = false
...
const processEvent = (time, args) => {
   if (overflow) return
   rows.push(...)
   overflow = rows.length >= maxCommandsPerRun
}
...
monitor.on("monitor", processEvent)
try {
   const end = Date.now() + monitoringDuration
   while (!overflow && Date.now() < end) await sleep(50)
} finally {
  try { await monitor.disconnect() } catch {}
  try { await client.disconnect() } catch {}
}

File: tools/valkey-metrics/src/effects/monitor-stream.js [hunk 10]
```diff
@@ -0,0 +1,57 @@
...
+    const overflow$ = new Subject()
+    
+    const processEvent = (time, args) => {
+      rows.push({ts: time, command: args.join(" ")})
+      if(rows.length >= maxLogs) overflow$.next()
+    }
+
+    monitor.on("monitor", processEvent)
+
+    let reason
```
arseny-kostenko: lets rename it to be more descriptive, like monitorCompletionReason

tools/valkey-metrics/src/effects/monitor-stream.js
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      reason = await firstValueFrom(
+        race([
+          timer(monitoringDuration).pipe(map(() => "Monitor duration completed.")),
+          overflow$.pipe(map(() => "Max logs read" )),
+        ])
+      )
+    } finally {
+      monitor.off("monitor", processEvent)
+      await monitor.disconnect()
```
arseny-kostenko: Promise.all them please

PR #59, opened 2025-11-04
Fix parsing bug

ravjotbrar:


apps/server/src/utils.ts
```diff
@@ -9,6 +9,11 @@ type ParsedClusterInfo = {
   }
 }
 
+export const parseResponse = (response: string) => {
```
arseny-kostenko: Ramda has cool helpers for situations like this:
export const parseResponse = (response: string) => R.when(R.includes(":"), parseInfo)(response)

or even one-liner:
export const parseResponse = R.when(R.includes(":"), parseInfo)

ravjotbrar: I like this better, thanks!

PR #77, opened 2025-11-19
Refactor monitoring under hotkeys endpoint

ravjotbrar:


tools/valkey-metrics/src/index.js
```diff
@@ -68,43 +68,63 @@ app.get('/slowlog_len', async (_req, res) => {
...
+    if(!monitorRunning) {
+      const result = await monitorHandler("start")
+      checkAt = result.checkAt
+      return res.json({checkAt})
+    }
+    const currentTime = Date.now()
+    const monitorCycle = monitorInterval + monitorDuration
+    if (currentTime > checkAt) {
+      const hotkeys = await calculateHotKeys()
+      checkAt = currentTime + monitorCycle
```
arseny-kostenko, line 124: looks like this won't stop the monitor and it will keep working

tools/valkey-metrics/src/index.js
```diff
@@ -68,43 +68,63 @@ app.get('/slowlog_len', async (_req, res) => {
...
-    const hotkeys = await calculateHotKeys()
-    res.json(hotkeys)
+    if(!monitorRunning) {
+      const result = await monitorHandler("start")
+      checkAt = result.checkAt
+      return res.json({checkAt})
+    }
+    const currentTime = Date.now()
+    const monitorCycle = monitorInterval + monitorDuration
+    if (currentTime > checkAt) {
```
arseny-kostenko, line 122: this code can be reached only when monitor is running BUT if the monitor is running, then calculateHotKeys will operate on an incomplete set of data

arseny-bitquill (approved):

PR #79, opened 2025-11-19
Integrate metrics into Electron app

ArgusLi:
## How is the metrics service integrated?

1. Electron app starts the server (no metric instances are started)
2. When the user clicks connect to an instance, electron app spins up a child process (CPX) with valkey-metrics for that connectionId.
3. CPX returns host and port of valkey-metrics instance to the Electron app.

## Notes
- Currently this does not support connecting to clusters, only standalone.

### DIFF
--- a/apps/frontend/src/state/epics/valkeyEpics.ts
+++ b/apps/frontend/src/state/epics/valkeyEpics.ts
@@ -21,7 +21,7 @@ import { connectFulfilled as wsConnectFulfilled } from "../wsconnection/wsConnec
 import history from "../../history.ts"
 import type { Store } from "@reduxjs/toolkit"
 
-export const connectionEpic = (store: Store) =>
+export const connectionEpic = () =>
   merge(
     action$.pipe(
       select(connectPending),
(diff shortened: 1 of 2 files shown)


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -28,21 +34,24 @@ export const connectionEpic = (store: Store) =>
...
-          R.pipe( // merge fulfilled connection with existing connections in localStorage
-            atId(connectionId),
-            R.evolve({ status: R.always(NOT_CONNECTED) }),
-            R.assoc(connectionId, R.__, currentConnections),
+          R.pipe(
+            (p) => ({
+              connectionDetails: p.connectionDetails,
+              status: NOT_CONNECTED,
+            }),
+            (connectionToSave) => R.assoc(payload.connectionId, connectionToSave, currentConnections),
```
arseny-kostenko: can be just
```
R.assoc(payload.connectionId, R.__, currentConnections),
```

https://ramdajs.com/docs/#__

don't have to change that. may as well just 
```
(connectionToSave) => ({ ...currentConnections, [payload.connectionId]: connectionToSave })
```

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -187,13 +195,20 @@ export const sendRequestEpic = () =>
 
 export const setDataEpic = () =>
   action$.pipe(
-    select(connectFulfilled),
+    filter(
+      (action) =>
```
arseny-kostenko: can be less verbose with destructuring action to ({ type })

arseny-bitquill (approved):

PR #80, opened 2025-11-20
Add query param for hotkeys

ravjotbrar:
1. by default hot-keys starts the monitor, runs it for one cycle, then stops the monitor
2. add a param to hot-keys that indicates that the user will be the one stopping the monitor; so the hot-keys endpoint will start the monitor;
3.  return data after one duration and keep the monitor running, and just always respond with the data. The user will have to call monitor stop themselves.
4. add a field to the response of hotkeys — monitorStatus


apps/metrics/src/index.js
```diff
@@ -108,34 +108,35 @@ async function main() {
   }
   app.get('/monitor', async (req, res) => {
     const result = await monitorHandler(req.query.action)
-    if (result.checkAt) checkAt = result.checkAt;
+    checkAt = result.checkAt
```
arseny-kostenko, line 113: since `checkAt` is a global-ish variable that's owned by `monitorHandler` — let's remove it from here and set it inside the `monitorHandler`

apps/metrics/src/index.js
```diff
@@ -108,34 +108,35 @@ async function main() {
...
   })
 
-  app.get('/hot-keys', async (_req, res) => {
+  app.get('/hot-keys', async (req, res) => {
+    let result = {}
     try {
       if (!monitorRunning) {
-        const result = await monitorHandler("start")
+        result = await monitorHandler(ACTION.START)
         checkAt = result.checkAt
```
arseny-kostenko: like above

apps/metrics/src/index.js
```diff
@@ -108,34 +108,35 @@ async function main() {
...
       }
-      const currentTime = Date.now()
-      const monitorCycle = monitorInterval + monitorDuration
-      if (currentTime > checkAt) {
+      if (Date.now() > checkAt) {
         const hotkeys = await calculateHotKeys()
-        checkAt = currentTime + monitorCycle
-        return res.json(hotkeys)
+        if(req.query.mode !== MODE.CONTINUOUS) {
+          result = await monitorHandler(ACTION.STOP)
```
arseny-kostenko: let's rename `result` to `monitorStatus` or something because the endpoint is `hot-keys`, so `result` is a very confusing name in such context

arseny-bitquill (approved):

PR #81, opened 2025-11-21
Pass metrics uri to backend

ArgusLi:
When a Metrics instance is created, we now pass the metrics server uri to the server app, where it is stored in a map of valkeyServerInstanceID to its respective metrics server URI.

On metrics instance close, we delete the metrics URI from the server.

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -16,13 +16,13 @@ const wss = new WebSocketServer({ port: 8080 })
 const metricsServerURIs: Map<string, string> = new Map()
 
 process.on("message", (message) => {
-  if (message && message.type === "metrics-started") {
+  if (message?.type === "metrics-started") {
     const metricsServerURI = `${message.payload.metricsHost}:${message.payload.metricsPort}`
     metricsServerURIs.set(message.payload.serverConnectionId, metricsServerURI)
     console.log(`Metrics server for ${message.payload.serverConnectionId} saved with URI ${metricsServerURI}`)
   }
 
-  if (message && message.type === "metrics-closed"){
+  if (message?.type === "metrics-closed"){
     if (metricsServerURIs.delete(message.payload.serverConnectionId)) {
       console.log(`Metrics server for ${message.payload.serverConnectionId} closed.`)
     }


apps/frontend/electron.main.js
```diff
@@ -51,12 +51,25 @@ function startMetrics(serverConnectionId, serverConnectionDetails) {
...
         }
     });
 
     metricsProcess.on('close', (code) => {
         console.log(`Metrics server for connection ${serverConnectionId} exited with code ${code}`);
         metricsProcesses.delete(serverConnectionId);
+        serverProcess.send({
+            type: 'metrics-closed',
+            payload: {
+                serverConnectionId: serverConnectionId
```
arseny-kostenko:
```
payload: {
  serverConnectionId,
}
```

apps/server/src/index.ts
```diff
@@ -13,6 +13,21 @@ import { setDashboardData, setClusterDashboardData } from "./setDashboardData.ts
 import { sendValkeyRunCommand } from "./sendCommand.ts"
 
 const wss = new WebSocketServer({ port: 8080 })
+const metricsServerURIs: Map<string, string> = new Map()
+
+process.on("message", (message) => {
+  if (message && message.type === "metrics-started") {
```
arseny-kostenko: can `message` be null | undefined actually?

apps/frontend/electron.main.js
ArgusLi: 🤦

apps/server/src/index.ts
ArgusLi: It can be undefined but not null.

arseny-kostenko: you can then write
```
if (message?.type === "metrics-started")

File: apps/frontend/electron.main.js [hunk 3]
```diff
@@ -51,12 +51,25 @@ function startMetrics(serverConnectionId, serverConnectionDetails) {
     metricsProcess.on('message', (message) => {
         if (message && message.type === 'metrics-started') {
             console.log(`Metrics server for ${serverConnectionId} started successfully on host: ${message.payload.metricsHost} port ${message.payload.metricsPort}`);
+            serverProcess.send?.({
```
ravjotbrar, line 54: Should this be process.post?

ArgusLi: It's confusing as there's both electron and node.js option. In this case, it's send as this is a node.js child process. https://nodejs.org/dist/latest-v16.x/docs/api/child_process.html#subprocesssendmessage-sendhandle-options-callback

PR #85, opened 2025-11-25
Integrate hotkeys endpoint with backend

ravjotbrar:


apps/server/src/actions/hotkeys.ts
```diff
@@ -0,0 +1,70 @@
...
+  hotkeys: [[]]
+  checkAt: number
+  monitorRunning: boolean
+}
+
+export const hotKeysRequested = withDeps<Deps, void>(
+  async ({ ws, connectionId, metricsServerURIs }) => {
+    const metricsServerURI = metricsServerURIs.get(connectionId)
+    try {
+      let response = await fetch(`${metricsServerURI}/hot-keys`)
```
arseny-kostenko: let's have two `const` and be explicit with the naming;

and add a comment with an explanation of how this works

arseny-kostenko: i.e. const per fetch

apps/server/src/actions/hotkeys.ts
```diff
@@ -0,0 +1,70 @@
...
+      let parsedResponse: HotKeysResponse = await response.json() as HotKeysResponse
+      console.log(parsedResponse)
+      if (parsedResponse.checkAt) {
+        const delay = Math.max(parsedResponse.checkAt - Date.now(), 0)
+        setTimeout(async () => {
+          try {
+            response = await fetch(`${metricsServerURI}/hot-keys`)
+            parsedResponse = await response.json() as HotKeysResponse
+            ws.send(
```
arseny-kostenko: let's extract them into two functions in this very file, so it's easier to read the logic
```
const sendHotKeysFulfilled = (
  ws: WebSocket,
  connectionId: string,
  parsedResponse: HotKeysResponse,
) => {
  ws.send(
    JSON.stringify({
      type: VALKEY.HOTKEYS.hotKeysFulfilled,
      payload: {
        connectionId,
        parsedResponse,
      },
    }),
  )
}

const sendHotKeysError = (
  ws: WebSocket,
  connectionId: string,
  error: unknown,
) => {
  console.log(error)
  ws.send(
    JSON.stringify({
      type: VALKEY.HOTKEYS.hotKeysError,
      payload: {
        connectionId,
        error: error instanceof Error ? error.message : String(error),
      },
    }),
  )
}
```

notice, I added `error instanceof Error ? error.message : String(error),` because not all json objects can be serialized — they can have [circular references, big integers which will make json.stringify throw an error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions)

install-mac-build.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+echo "🛑 Closing existing Skyscope app if running…"
```
arseny-kostenko: let's reduce the number of emojis and keep only terminal states — installed and launched; and errors

arseny-bitquill (approved):

PR #86, opened 2025-11-28
Get hotkeys for all cluster nodes

ravjotbrar:


apps/frontend/electron.main.js
```diff
@@ -20,6 +20,20 @@ function startServer() {
     }
 }
 
+function startMetricsForClusterNodes(clusterNodes) {
```
ArgusLi, line 23: Do we want to store a map of clusterIDs to connectionIds?

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -251,11 +251,25 @@ export const setDataEpic = () =>
...
     select(hotKeysRequested),
     tap((action) => {
+      const { clusterId, connectionId } = action.payload
       const socket = getSocket()
-      socket.next(action)
+
+      const state = store.getState()
+      const clusters = state.valkeyCluster.clusters
+
+      const connectionIds =
```
ArgusLi, line 265: Should we be storing this info here? Or in the Electron app?

ravjotbrar: Good question. I think until I have good reason to change it, we'll keep it as is.

apps/frontend/electron.main.js
arseny-kostenko: ~agree, I'd maintain a map with statuses~

apps/frontend/src/state/epics/valkeyEpics.ts
arseny-kostenko: we definitely need a place in redux state to track connectionId: { status: "requested" | "fulfilled", ts }

arseny-kostenko: without it we can't show in UI that the request is pending

apps/server/src/connection.ts
```diff
@@ -149,15 +149,19 @@ async function connectToCluster(
 
   clients.set(payload.connectionId, clusterClient)
 
+  const clusterConnectionInfo = {
+    type: VALKEY.CONNECTION.clusterConnectFulfilled,
+    payload: {
+      connectionId: payload.connectionId,
+      clusterNodes,
+      clusterId: clusterId,
```
arseny-kostenko: `clusterId,`

apps/frontend/src/state/epics/valkeyEpics.ts
ravjotbrar: Added status in hotKeys state

arseny-bitquill (approved):

PR #87, opened 2025-11-28
implemented functionality of slow logs from server to frontend

nassery318:


apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+  }, [id, dispatch])
+
+  const tabs = [
+    { id: "hot-keys" as TabType, label: "Hot Keys" },
+    { id: "large-keys" as TabType, label: "Large Keys" },
+    { id: "slow-logs" as TabType, label: "Slow Logs" },
+  ]
+
+  const toggleSortOrder = () => {
+    setSortOrder((prev) => prev === "asc" ? "desc" : "asc")
```
ravjotbrar, line 39: Doesn't have to be in this PR, but it might be worth sorting by the slowest, largest, and hottest keys for each column.

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+import type { RootState } from "@/store"
+import { slowLogsRequested, selectSlowLogs } from "@/state/valkey-features/slowlogs/slowLogsSlice"
+import { useAppDispatch } from "@/hooks/hooks"
+
+type TabType = "hot-keys" | "large-keys" | "slow-logs"
+type SortOrder = "asc" | "desc"
+
+export const Monitoring = () => {
+  const dispatch = useAppDispatch()
+  const { id } = useParams()
```
ravjotbrar, line 16: You're likely going to need clusterId param here too when dispatching actions for a cluster node. Can do that in your hotkeys PR.

apps/frontend/src/components/Monitoring.tsx
nassery318: Yes for sure, once the hot keys functionality is sorted out, I will be refactoring this whole component.

apps/frontend/src/components/Monitoring.tsx
nassery318: Yes will be done in the next PR as hot keys are not part of this PR.

ravjotbrar: I'm assuming we'll be fetching slowlogs and bigkeys at a cluster level too, but yeah we don't need it until your hotkeys PR.

nassery318: Yes

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+    setSortOrder((prev) => prev === "asc" ? "desc" : "asc")
+  }
+
+  const sortedSlowLogs = [...(slowLogsData || [])].flatMap((logGroup) =>
+    logGroup.values.map((entry) => ({
+      ...entry,
+      groupTs: logGroup.ts,
+    })),
+  ).sort((a, b) => {
+    const timestampA = new Date(a.ts).getTime()
```
arseny-kostenko: destructuring above is unnecessary — it's allocating a new array for the same data

you already have ts as a number, you don't need to create a Date from it and then again get the same value, i.e. 
```
a.ts === new Date(a.ts).getTime()
```

can rewrite much 
```
const sortedSlowLogs = R.pipe(
  R.defaultTo([]),
  R.chain((logGroup) =>
    logGroup.values.map((entry) => ({
      ...entry,
      groupTs: logGroup.ts,
    }),
  ),
  R.sort(sortOrder === "asc" ? R.ascend(R.prop("ts")) : R.descend(R.prop("ts"))),
)(slowLogsData)
```

apps/frontend/src/components/Monitoring.tsx
arseny-kostenko: My understand is that once broken down into separate components hotkeys, slowlog, largekeys components — they each will be per connectionId, but a cluster level component will be just a container for all connectionIds, i.e. cluster.nodes.map(hotkeys), cluster.nodes.map(largekeys) etc;

apps/frontend/src/components/Monitoring.tsx
arseny-kostenko: at least, you can replace `[...(slowLogsData || [])]` with `R.defaultTo([], slowLogsData)`

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <div className="h-full flex items-center justify-center">
+            <span className="text-lg text-gray-500 dark:text-white mb-2">
+              No Large Keys Found!
+            </span>
+          </div>
+        )}
+
+        {activeTab === "slow-logs" && (
+          <div className="h-full w-full flex flex-col">
+            {sortedSlowLogs?.length > 0 ? (
```
arseny-kostenko: `sortedSlowLogs` can't be null — it's default to `[]`

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+                        </span>
+                      </div>
+                    ))}
+                  </div>
+                </div>
+              </div>
+            ) : (
+              <div className="h-full flex items-center justify-center">
+                <span className="text-lg text-gray-500 dark:text-white mb-2">
+                  No Slow Logs Found!
```
arseny-kostenko: let's remove exclamation mark

apps/frontend/src/state/valkey-features/slowlogs/slowLogsSlice.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const initialSlowLogsState: SlowLogsState = {}
+
+const slowLogsSlice = createSlice({
+  name: "slowLogs",
+  initialState: initialSlowLogsState,
+  reducers: {
+    slowLogsRequested: (state, action) => {
+      const connectionId = action.payload.connectionId
+      if (!state[connectionId]) {
```
arseny-kostenko: don't have to do it — can collapse this using ??=

```
(state[connectionId] ??= {
  slowLogs: [],
  count: 50,
  loading: false,
}).loading = true
```

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+          </button>
+        )}
+      </div>
+
+      {/* Tab Content */}
+      <div className="flex-1 border dark:border-tw-dark-border rounded overflow-y-auto">
+        {activeTab === "hot-keys" && (
+          <div className="h-full flex items-center justify-center">
+            <span className="text-lg text-gray-500 dark:text-white mb-2">
+              No Hot Keys Found!
```
arseny-kostenko: let's remove the exclamation mark

apps/frontend/src/components/Monitoring.tsx
```diff
@@ -0,0 +1,178 @@
...
+            </span>
+          </div>
+        )}
+        {activeTab === "large-keys" && (
+          <div className="h-full flex items-center justify-center">
+            <span className="text-lg text-gray-500 dark:text-white mb-2">
+              No Large Keys Found!
```
arseny-kostenko: let's remove the exclamation mark

arseny-bitquill (approved):

PR #90, opened 2025-12-01
added hot keys to the monitoring view and refactored the monitoring page

nassery318:


apps/frontend/src/components/ui/hot-keys.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function HotKeys({ data }: HotKeysProps) {
+  const [sortOrder, setSortOrder] = useState<SortOrder>("desc")
+
+  const toggleSortOrder = () => {
+    setSortOrder((prev) => prev === "asc" ? "desc" : "asc")
+  }
+
+  const sortedHotKeys = [...R.defaultTo([], data)].sort((a, b) => {
```
arseny-kostenko: you don't need destructuring here

apps/frontend/src/components/ui/hot-keys.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export function HotKeys({ data, status }: HotKeysProps) {
+  const [sortOrder, setSortOrder] = useState<SortOrder>("desc")
+
+  const toggleSortOrder = () => {
+    setSortOrder((prev) => prev === "asc" ? "desc" : "asc")
+  }
+
+  const sortedHotKeys = R.sort<[string, number]>(
+    (a, b) => {
```
arseny-kostenko: since you chose ramda's sort, you can write it more concisely:
```
R.sort<[string, number]>((sortOrder === "asc" ? R.ascend : R.descend)(R.nth(1)))
```

apps/frontend/src/components/ui/slow-logs.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const toggleSortOrder = () => {
+    setSortOrder((prev) => prev === "asc" ? "desc" : "asc")
+  }
+
+  const sortedSlowLogs = R.defaultTo([], data).flatMap((logGroup) =>
+    logGroup.values.map((entry) => ({
+      ...entry,
+      groupTs: logGroup.ts,
+    })),
+  ).sort((a, b) => {
```
arseny-kostenko: if you want, same way as above, you can write it in a less verbose way
```
.sort((sortOrder === "asc" ? R.ascend : R.descend)(R.prop("ts")))
  ```

nassery318: addressed

apps/frontend/src/components/ui/hot-keys.tsx
nassery318: addressed

PR #91, opened 2025-12-01
Fix Fluctuating key count

ArgusLi:
GetKeys() gets all keys for each node until cursor is 0.
When scanning a cluster, it gets all key for each node until cursor is 0 in parallel.


apps/frontend/src/components/ui/key-details.tsx
```diff
@@ -46,10 +46,10 @@ interface keyDetailsProps {
   selectedKey: string | null;
   setSelectedKey: (key: string | null) => void;
   selectedKeyInfo: KeyInfo | null;
-  conectionId: string;
+  connectionId: string;
 }
 
-export default function KeyDetails({ selectedKey, selectedKeyInfo, conectionId, setSelectedKey }: keyDetailsProps) {
+export default function KeyDetails({ selectedKey, selectedKeyInfo, connectionId: conectionId, setSelectedKey }: keyDetailsProps) {
```
arseny-kostenko: so you almost fixed the typo?

ArgusLi: 🤦‍♂️

apps/frontend/src/components/ui/key-details.tsx
```diff
@@ -46,10 +46,10 @@ interface keyDetailsProps {
   selectedKey: string | null;
   setSelectedKey: (key: string | null) => void;
   selectedKeyInfo: KeyInfo | null;
-  conectionId: string;
+  connectionId: string;
 }
 
-export default function KeyDetails({ selectedKey, selectedKeyInfo, conectionId, setSelectedKey }: keyDetailsProps) {
+export default function KeyDetails({ selectedKey, selectedKeyInfo, connectionId: connectionId, setSelectedKey }: keyDetailsProps) {
```
ravjotbrar:
```suggestion
export default function KeyDetails({ selectedKey, selectedKeyInfo, connectionId, setSelectedKey }: keyDetailsProps) {
```

apps/server/src/keys-browser.ts
```diff
@@ -82,37 +82,123 @@ export async function getKeyInfo(
...
+    pattern?: string;
+    count?: number;
+  }, 
+): Promise<Set<string>> {
+  const pattern = payload.pattern || "*"
+  const count = payload.count || 50 // this is JUST A HINT. may return larger than count https://valkey.io/commands/scan/
+  const routeOption: RouteOption = { route: "allPrimaries" } // Sends command to all primary nodes
+  const allKeys = new Set<string>()
+    
+  const scanClusterResult = (await client.customCommand([
```
ravjotbrar: We could pull this custom command out into its own function since we repeat it

apps/server/src/keys-browser.ts
```diff
@@ -82,37 +82,123 @@ export async function getKeyInfo(
...
+    "COUNT",
+    count.toString(),
+  ],
+  routeOption,
+  )) as ClusterScanResult[]
+
+  console.log("SCAN Cluster response:", scanClusterResult)
+
+  await Promise.all(scanClusterResult.map(async ({ key: nodeAddress, value })=>{
+    value[1].forEach((nodeKey) => {
```
ravjotbrar: Might be worth clarifying what value[1] is

apps/server/src/keys-browser.ts
```diff
@@ -82,37 +82,123 @@ export async function getKeyInfo(
...
       "COUNT",
       count.toString(),
-    ])) as [string, string[]]
+    ],
+    )) as [string, string[]]
+
+    console.log("SCAN standalone response:", scanResult)
+
+    cursor = scanResult[0]
+    scanResult[1].forEach((key) => {allKeys.add(key)})
```
ravjotbrar: is there no limit to how many keys we retrieve?

apps/frontend/src/components/ui/key-details.tsx
ArgusLi: Addressed

apps/server/src/keys-browser.ts
ArgusLi: Doesn't the `.forEach((nodeKey))` state that value[1] is a list of keys?

apps/server/src/keys-browser.ts
ArgusLi: We don't limit currently.

ArgusLi: Count is a hint for the rough number of keys return per call in SCAN.
https://valkey.io/commands/scan/

apps/server/src/keys-browser.ts
ArgusLi: This is a surprisingly difficult custom command to modularize. We receive different result schemas based on if it's a GlideClient or GlideClusterClient, and how we loop it differs if we are using GlideClient or GlideClusterClient. There's also not that much code to modularize, so it would look very similar to the code now.

It seems to me to be adding complexity without much gain.

ArgusLi: I will set all magic numbers to a single source of reference.

apps/server/src/keys-browser.ts
ravjotbrar: I thought we were going to limit to 1000?

apps/server/src/keys-browser.ts
ravjotbrar: If it's a lot of complexity to pull out into its own function, then sure we don't need to. But we repeat a similar command three times, so it would make the code easier to read.

apps/server/src/keys-browser.ts
ArgusLi: That's a future ticket and will involve UI changes.

apps/server/src/keys-browser.ts
ArgusLi: It will increase complexity while making it more difficult to read. The Scan custom commands have different return signatures which are not segregated by Glide Client type. It would be easy to use the wrong function or read it and misunderstand the function.

apps/server/src/keys-browser.ts
```diff
@@ -82,37 +82,125 @@ export async function getKeyInfo(
...
+
+async function scanCluster(
+  client: GlideClusterClient, 
+  payload: {
+    connectionId: string;
+    pattern?: string;
+    count?: number;
+  }, 
+): Promise<Set<string>> {
+  const pattern = payload.pattern || VALKEY_CLIENT.SCAN.defaultPayloadPattern
```
arseny-kostenko:
```
const {
  pattern = VALKEY_CLIENT.SCAN.defaultPayloadPattern,
  count = VALKEY_CLIENT.SCAN.defaultCount,
} = payload
```
embrace destructuring with default valeus

apps/server/src/keys-browser.ts
arseny-kostenko: what Ravjot is saying is that we repeat 3 times this:
```
[
    "SCAN",
    "0",
    "MATCH",
    pattern,
    "COUNT",
    count.toString(),
  ]
  ```
  
  which could be extracted into a separate module with commands. And by commands I mean indeed `[COMMAND, ...rest of its paramteres]` without client dependant things. So I agree with him.

Comment (ArgusLi) [reply apps/server/src/keys-browser.ts]: So like this right?

```
async function scan(
  client: GlideClient | GlideClusterClient,
  cursor: string,
  pattern?: string,
  count?: number,
  routeOption?: RouteOption,
): Promise<[string, string[]]> {
  const commandArgs = 
    [
      "SCAN", 
      cursor, 
      "MATCH", 
      pattern ||  VALKEY_CLIENT.SCAN.defaultPayloadPattern, 
      "COUNT", 
      count ? count.toString() : VALKEY_CLIENT.SCAN.defaultCount,
    ]

  if (client instanceof GlideClusterClient) {
    return (await client.customCommand(commandArgs, routeOption)) as [string, string[]]
  } else {
    return (await client.customCommand(commandArgs)) as [string, string[]]
  }
}
```

Comment (ArgusLi) [reply apps/server/src/keys-browser.ts]: Then we would also need a second one for the cluster so something like this
```
async function scanClusterNodes(
  client: GlideClusterClient,
  cursor: string,
  pattern?: string,
  count?: number,
  routeOption?: RouteOption,
): Promise<ClusterScanResult[]> {
  const commandArgs = 
    [
      "SCAN", 
      cursor, 
      "MATCH", 
      pattern ||  VALKEY_CLIENT.SCAN.defaultPayloadPattern, 
      "COUNT", 
      count ? count.toString() : VALKEY_CLIENT.SCAN.defaultCount,
    ]

  return (await client.customCommand(commandArgs, routeOption)) as ClusterScanResult[]
}
```

File: apps/server/src/keys-browser.ts [hunk 7]
```diff
@@ -82,37 +82,125 @@ export async function getKeyInfo(
...
+  )) as ClusterScanResult[]
+
+  console.log("SCAN Cluster response:", scanClusterResult)
+
+  await Promise.all(scanClusterResult.map(async ({ key: nodeAddress, value })=>{
+    value[1].forEach((nodeKey) => {
+      allKeys.add(nodeKey)
+    })
+
+    const routeByAddress: SingleNodeRoute = {
```
arseny-kostenko: it's verbose as it's splitting it twice:
```
const routeByAddress: SingleNodeRoute = R.pipe(
  R.split(":"),
  ([host, port]) => ({
      type: "routeByAddress",
      host,
      port: Number(port),
  }),
)(nodeAddress)
```

apps/server/src/keys-browser.ts
ArgusLi: Or do you mean we store a set of commandArgs, and use it in the function?
So something like
```
const buildCommandArgs(cursor, pattern?, count?) => {
    return [
      "SCAN", 
      cursor, 
      "MATCH", 
      pattern ||  VALKEY_CLIENT.SCAN.defaultPayloadPattern, 
      "COUNT", 
      count ? count.toString() : VALKEY_CLIENT.SCAN.defaultCount,
    ]}
```

apps/server/src/keys-browser.ts
ArgusLi: How is Typescript getting an error from this...
```
Type '{ type: string; host: string; port: number; }' is not assignable to type 'SingleNodeRoute'.
  Type '{ type: string; host: string; port: number; }' is not assignable to type 'SlotIdTypes | SlotKeyTypes | RouteByAddress'.
    Type '{ type: string; host: string; port: number; }' is not assignable to type 'RouteByAddress'.
      Types of property 'type' are incompatible.
        Type 'string' is not assignable to type '"routeByAddress"'.ts(2322)
```

arseny-kostenko: either turn string type to a literal type by doing
```
type: "routeByAddress" as const,
```
or annotate the entire return value 
```
([host, port]): SingleNodeRoute => ({
```
— you probably don't need both

apps/server/src/keys-browser.ts
arseny-kostenko: yes, the last one. And use object instead of overloading
```
const buildCommandArgs = ({
  cursor,
  pattern = VALKEY_CLIENT.SCAN.defaultPayloadPattern,
  count = VALKEY_CLIENT.SCAN.defaultCount,
}) => ["SCAN", cursor, "MATCH", pattern, "COUNT", String(count)]
```

apps/server/src/keys-browser.ts
arseny-kostenko: you can destructure it into a meaningful name in the map fn signature:
```
map(async ({ key: nodeAddress, [, nodeKeys] })=>{
```

apps/server/src/keys-browser.ts
```diff
@@ -82,37 +82,125 @@ export async function getKeyInfo(
...
+  }, 
+): Promise<Set<string>> {
+  const pattern = payload.pattern || VALKEY_CLIENT.SCAN.defaultPayloadPattern
+  // count is JUST A HINT. may return larger than count https://valkey.io/commands/scan/
+  const count = payload.count || VALKEY_CLIENT.SCAN.defaultCount
+  const allKeys = new Set<string>()
+    
+  let cursor = "0"
+  do {
+    const scanResult = (await client.customCommand([
```
arseny-kostenko: you can destructure scanResult to use meaningful names down there
```
const [cursor, keys] = (await client.customCommand([
...
```
to avoid mysterious scanResult[1]

apps/server/src/keys-browser.ts
ArgusLi: That works! Let me look into why that works...

apps/server/src/valkey-client-commands.ts
```diff
@@ -0,0 +1,13 @@
+import { VALKEY_CLIENT } from "../../../common/src/constants.ts"
```
arseny-kostenko, line 1: the file named using the wrong convention, please rename it as kebab space to valkey-client-commands.ts as per node.js convention

PR #93, opened 2025-12-02
Make slowlogs cluster compatible

ravjotbrar:


apps/metrics/src/index.js
```diff
@@ -54,9 +54,13 @@ async function main() {
...
   app.get("/slowlog", async (req, res) => {
     try {
-      const count = Number(req.query.count) || 50
-      const rows = await Streamer.slowlog_get(count)
-      res.json({ count: Math.max(1, Math.min(500, count)), rows })
+      const {lastUpdatedAt, nextCycleAt} = getCollectorMeta(SLOWLOG)
+      if (lastUpdatedAt !== null) {
+        const count = Number(req.query.count) || 50
+        const rows = await Streamer.slowlog_get(count)
+        return res.json({ count: Math.max(1, Math.min(500, count)), rows, lastUpdatedAt })
```
arseny-kostenko, line 62: can you leave a comment in the code as to why `Math.max(1, ..)`? — I'm confused

apps/server/src/actions/slowLogs.ts
```diff
@@ -52,25 +53,38 @@ const sendSlowLogsError = (
...
 export const slowLogsRequested = withDeps<Deps, void>(
-  async ({ ws, connectionId, metricsServerURIs }) => {
-    const metricsServerURI = metricsServerURIs.get(connectionId)
+  async ({ ws, metricsServerURIs, action }) => {
+    const { connectionIds } = action.payload
+    const promises = connectionIds.map(async (connectionId: string) => {
+      const metricsServerURI = metricsServerURIs.get(connectionId)
+      try {
+        const count = 50 
+        const url = `${metricsServerURI}/slowlog?count=${count}`
```
arseny-kostenko: we don't have to specify `count` as it defaults to 50 anyways in the metrics servce

apps/server/src/actions/slowLogs.ts
```diff
@@ -52,25 +53,38 @@ const sendSlowLogsError = (
...
+        console.log("[slowLogsRequested] Fetching from:", url)
 
-    try {
-      const count = 50 
-      const url = `${metricsServerURI}/slowlog?count=${count}`
-      console.log("[slowLogsRequested] Fetching from:", url)
+        const initialResponse = await fetch(url)
+        const initialParsedResponse: SlowLogsResponse = await initialResponse.json() as SlowLogsResponse
+        if (initialParsedResponse.checkAt) {
+          const delay = Math.max(initialParsedResponse.checkAt - Date.now(), 0)
```
arseny-kostenko: you don't have to Math.max — setTimeout can accept negative values which means "now"

apps/metrics/src/index.js
ravjotbrar: Not sure, I didn't write the original code, but it seems like it enforces a maximum row count of 500 and minimum of 1.

apps/server/src/actions/slowLogs.ts
ravjotbrar: Fair. I'm assuming we'll eventually make this configurable anyways.

apps/server/src/actions/slowLogs.ts
ravjotbrar: Changed

PR #98, opened 2025-12-06
Refactor Slowlogs into CommandLogs

ravjotbrar:
Also added eslint rules for js files, but won't be fixing all eslint issues to minimize the amount of changes in this PR. Will create a new PR to fix all linting.

### DIFF
--- a/apps/metrics/src/effects/ndjson-streamer.js
+++ b/apps/metrics/src/effects/ndjson-streamer.js
@@ -12,7 +12,7 @@ const fileFor = (prefix, date) => {
   return path.join(dataDir, `${prefix}_${dayStr(date)}.ndjson`)
 }
 
-export async function streamNdjson(prefix, filterFn = () => true, limit = Infinity) {
+export async function streamNdjson(prefix,{ filterFn = () => true, limit = Infinity } = {}) {
   const today = new Date()
   const yesterday = new Date(today)
   yesterday.setDate(today.getDate() - 1)
@@ -52,4 +52,4 @@ export async function streamNdjson(prefix, filterFn = () => true, limit = Infini
 
 export const [memory_stats, info_cpu, slowlog_len, commandlog_slow, commandlog_large_reply, commandlog_large_request, monitor] =
   // eslint-disable-next-line max-len
-  ["memory", "cpu", "slowlog_len", COMMANDLOG_SLOW, COMMANDLOG_LARGE_REPLY, COMMANDLOG_LARGE_REQUEST, MONITOR].map((filePrefix) => () => streamNdjson(filePrefix))
+  ["memory", "cpu", "slowlog_len", COMMANDLOG_SLOW, COMMANDLOG_LARGE_REPLY, COMMANDLOG_LARGE_REQUEST, MONITOR].map((filePrefix) => (options = {}) => streamNdjson(filePrefix, options))
--- a/apps/metrics/src/handlers/commandlog-handler.js
+++ b/apps/metrics/src/handlers/commandlog-handler.js
@@ -5,11 +5,11 @@ const getCommandLogRows = async (commandlogType, count) => {
   try {
     switch (commandlogType) {
       case COMMANDLOG_TYPE.SLOW:
-        return Streamer.commandlog_slow(count)
+        return Streamer.commandlog_slow({ limit: count })
       case COMMANDLOG_TYPE.LARGE_REQUEST:
-        return Streamer.commandlog_large_request(count)
+        return Streamer.commandlog_large_request({ limit: count })
       case COMMANDLOG_TYPE.LARGE_REPLY:
-        return Streamer.commandlog_large_reply(count)
+        return Streamer.commandlog_large_reply({ limit: count })
       default:
         throw new Error(`Unknown commandlog type: ${commandlogType}`)
     }


apps/metrics/src/utils/helpers.js
```diff
@@ -1,6 +1,20 @@
-export const ymd = d => {
+export const ymd = (d) => {
   const y = d.getFullYear()
   const m = String(d.getMonth() + 1).padStart(2, "0")
   const day = String(d.getDate()).padStart(2, "0")
   return `${y}${m}${day}` // "20250924"
 }
+
+export const parseCommandLogs = (entries) => {
```
ArgusLi: We can destructure in one go like this:
```js
export const parseCommandLogs = (entries) => 
  entries.map(([
    id, 
    tsSec, 
    durationUs, 
    argv = [], 
    addr = "unknown", 
    name = "unknown"
  ] = []) => ({
    id: String(id),
    ts: Number(tsSec) * 1000,
    duration_us: Number(durationUs),
    argv: Array.isArray(argv) ? argv.map(String) : [],
    addr: String(addr),
    client: String(name),
  }));
```

apps/metrics/src/utils/constants.js
```diff
@@ -9,4 +11,10 @@ export const ACTION = {
   START: "start",
   STOP: "stop",
   STATUS: "status",
-}
\ No newline at end of file
+}
+
+export const COMMANDLOG_TYPE = {
```
ArgusLi, line 18: This seems to be duplicated in `common/src/constants.ts`. Is there a reason for these constants (and those above) to be in `apps/metrics/src/utils/constants.js`?

apps/metrics/src/index.js
```diff
@@ -77,63 +66,26 @@ async function main() {
...
   app.get("/monitor", async (req, res) => {
     const result = await monitorHandler(req.query.action)
-    checkAt = result.checkAt
     return res.json(result)
   })
 
   app.get("/hot-keys", async (req, res) => {
     let monitorResponse = {}
+    const { isRunning, willCompleteAt } = getCollectorMeta(MONITOR) 
+    const checkAt = willCompleteAt
```
ArgusLi: Can we rename the variable in the earlier line?
```js
const { isRunning, willCompleteAt: checkAt } = getCollectorMeta(MONITOR);
```

apps/metrics/src/init-collectors.js
```diff
@@ -22,49 +22,51 @@ const updateCollectorMeta = (name, patch) => {
...
+    appendRows: async (rows) => {
       await nd.appendRows(rows)
       console.info(`[${monitorEpic.name}] wrote ${rows.length} logs to ${cfg.server.data_dir}/`)
     },
-    close: nd.close
+    close: nd.close,
   }
-
-  const now = Date.now()
+  const monitorDuration = monitorEpic.monitoringDuration
```
ArgusLi: Should we just get rid of this constant and use it directly? It is only used once.

apps/metrics/src/effects/ndjson-streamer.js
```diff
@@ -38,8 +38,9 @@ export async function streamNdjson(prefix, filterFn = () => true) {
...
-  return results; // JSON array of all objects
+  return results // JSON array of all objects
 }
 
-export const [memory_stats, info_cpu, slowlog_len, slowlog_get, monitor] =
-  ['memory', 'cpu', 'slowlog_len', 'slowlog', 'monitor'].map(filePrefix => () => streamNdjson(filePrefix))
\ No newline at end of file
+export const [memory_stats, info_cpu, slowlog_len, commandlog_slow, commandlog_large_reply, commandlog_large_request, monitor] =
+  // eslint-disable-next-line max-len
+  ["memory", "cpu", "slowlog_len", COMMANDLOG_SLOW, COMMANDLOG_LARGE_REPLY, COMMANDLOG_LARGE_REQUEST, MONITOR].map((filePrefix) => () => streamNdjson(filePrefix))
```
ArgusLi: Should slowlog_len also be coming from the constants file? Also, looking at e.g. `COMMANDLOG_SLOW`, while it's added to the constants file, there are a lot of usages referring directly to the string "commandlog_slow" instead of the constant. Would this be the same for the other constants?

ArgusLi: ndjson-streamer is called from commandlog-handler with e.g. `Streamer.commandlog_slow(count)`. Looking at this export, the count is just dumped. Is there a need for count (as it seems to be somewhat arbitrary)? Can we just set a default value here for number of rows?

apps/server/src/actions/commandLogs.ts
```diff
@@ -0,0 +1,113 @@
...
+    }),
+  )
+}
+
+export const commandLogsRequested = withDeps<Deps, void>(
+  async ({ ws, metricsServerURIs, action }) => {
+    const { connectionIds } = action.payload
+    const commandLogType: CommandLogType = action.payload.commandLogType as CommandLogType
+    const promises = connectionIds.map(async (connectionId: string) => {
+      const metricsServerURI = metricsServerURIs.get(connectionId)
```
ArgusLi, line 83: This does not work for clusters right? Or are connectionIds in the action.payload for each cluster primary node already?

apps/metrics/src/utils/constants.js
ravjotbrar: I didn't want to spend too much time trying to get them to resolve in the metrics app, so I just duplicated a file for now. I ran into some issues before with getting common to work in metrics

apps/server/src/actions/commandLogs.ts
ravjotbrar: The latter

apps/metrics/src/effects/ndjson-streamer.js
ravjotbrar: Usages other than the ones in the config were missed. I'll change those.

apps/metrics/src/utils/constants.js
ArgusLi: Ah, makes sense. Then could you please add a comment here?

apps/metrics/src/effects/ndjson-streamer.js
ravjotbrar: Good catch, let me dig into this a bit deeper.

ravjotbrar: I've added logic to actually use the count as a limit for the number of rows we stream. It'll default to infinity, but for commandlogs we'll default to 50 for now.

ArgusLi: I believe you missed this line. Should be like this right?
`.map((filePrefix) => (limit) => streamNdjson(filePrefix, undefined, limit));`

ravjotbrar: Good catch again. Changed to options.

ArgusLi: Making it options is even better. Nice!

PR #101, opened 2025-12-09
Fix notarize.js if env file does not exist

ArgusLi:
- Disable notarization if .env file does not exist.
- Add note in read me about cluster reset.

### DIFF
--- a/mac_build/notarize.js
+++ b/mac_build/notarize.js
@@ -1,25 +1,26 @@
 /* eslint-disable @typescript-eslint/no-require-imports */
 const { notarize } = require("electron-notarize")
 
-function isEnvFilePresent() {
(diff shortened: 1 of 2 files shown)


mac_build/notarize.js
```diff
@@ -1,18 +1,26 @@
...
+    return false
+  }
+
+  return true
 }
 
 exports.default = async function notarizing(context) {
-  if (!result || process.env.SKIP_NOTARIZE === "true") {
-    console.log("  • No mac_build/.env or SKIP_NOTARIZE=true detected. Skipping notarization step.")
+  if (!isEnvFilePresent()) {
```
ravjotbrar: Is there a way to just skip notarization by detecting nosign in `npm run package:mac:nosign`?

ArgusLi: You are right. I have addressed this.

PR #104, opened 2025-12-10
added sorting based on size and duration for command logs

nassery318:


apps/frontend/src/components/ui/command-log-table.tsx
```diff
@@ -3,6 +3,7 @@ import { ArrowUp, ArrowDown, Clock } from "lucide-react"
 import * as R from "ramda"
 
 type SortOrder = "asc" | "desc"
```
ravjotbrar: Let's make asc and desc into consts

nassery318: addressed

ravjotbrar: Sorry I should've been more clear. In the common/src/constants file so they are all contained there.

PR #105, opened 2025-12-11
Add slot-level approach for getting hotkeys

ravjotbrar:
Pre-reqs:
`CONFIG SET maxmemory-policy allkeys-lfu`
`CONFIG set cluster-slot-stats-enabled yes`

Both of these are now set in our cluster_init script. 

Still need to test more keys to see if the approach is correct. We need to fix how we get keys to avoid sending too many inflight requests at once.

### DIFF
--- a/apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
+++ b/apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
@@ -13,6 +13,7 @@ interface ConnectionDetails {
   alias?: string;
   role?: Role;
   clusterId?: string;
+  // Eviction policy required for getting hot keys using hot slots
   lfuEnabled?: boolean;
 }
 
--- a/apps/metrics/src/analyzers/calculate-hot-keys.js
+++ b/apps/metrics/src/analyzers/calculate-hot-keys.js
@@ -26,14 +26,12 @@ export const calculateHotKeysFromHotSlots = async (client, count = 50) => {
   const hotSlots = await getHotSlots(client)
 
   const slotPromises = hotSlots.map(async (slot) => {
-    //console.log("Slot is: ", slot)
     const slotId = slot["slotId"]
     const keys = []
     let cursor = slotId
     let cursorToSlot = slotId
 
     do {
-      //console.log("Cursor is: ", cursor)
       const [nextCursor, scannedKeys] = await client.sendCommand(["SCAN", cursor.toString(), "COUNT", "1"])
       cursor = nextCursor
       keys.push(...scannedKeys)
@@ -42,6 +40,7 @@ export const calculateHotKeysFromHotSlots = async (client, count = 50) => {
     } while (cursorToSlot === slotId && cursor !== 0)
     
     return keys.map( async (key) => {
+      // Must have LFU enabled for this to work
       const freq = parseInt(await client.sendCommand(["OBJECT", "FREQ", key]))
       return { key, freq }
     })


apps/metrics/src/handlers/lfu-hotkeys-handler.js
```diff
@@ -0,0 +1,15 @@
+import { ALLKEYS_LFU, VOLATILE_LFU } from "../utils/constants"
+
+export const evictionPolicyIsLFU = async (client) => {
```
ArgusLi: nit: can we rename this isEvictionPolicyLFU?

apps/metrics/src/handlers/lfu-hotkeys-handler.js
```diff
@@ -0,0 +1,15 @@
...
+  const result = await client.sendCommand(["MEMORY", "STATS"])
+  return result === VOLATILE_LFU || result === ALLKEYS_LFU
+}
+function clusterEnabled(info) {
```
ArgusLi: nit: can we rename this isClusterEnabled?

apps/server/src/keys-browser.ts
```diff
@@ -154,8 +154,6 @@ async function scanCluster(
         nodeRouteOption,
       )) as [string, string[]]
 
-      console.log("SCAN node response:", scanResult)
```
ArgusLi, line 157: Why remove this?

tools/valkey-cluster/populate.mjs
```diff
@@ -39,13 +35,32 @@ async function main() {
     ...I.map(i => cluster.setBit("bitmap", i, 1)),
   ])
 
-  // streams gotta be sequential
   for (let i = 1; i <= 5; i++) {
     await cluster.xAdd("stream", "*", { sensor: `${1000 + i}`, value: `${20 + i}` })
-    await sleep(50) // sleep here is to make each xAdd get a unique, increasing timestamp
```
ArgusLi, line 45: Why remove this comment?

tools/valkey-cluster/populate.mjs
```diff
@@ -39,13 +35,32 @@ async function main() {
...
+    for (let i = start; i <= batchEnd; i++) {
+      promises.push(cluster.set(`bulk:key:${i}`, `value_${i}`))
+    }
+
+    await Promise.all(promises)
+    console.log(`Created keys ${start} → ${batchEnd}`)
   }
 
-  console.log("Loaded entries for all Valkey data types.")
+  console.log("All 35,000 additional string keys created successfully.")
```
ArgusLi: Can we use TOTAL_KEYS for this value?

tools/valkey-cluster/populate.mjs
```diff
@@ -39,13 +35,32 @@ async function main() {
...
     await cluster.xAdd("stream", "*", { sensor: `${1000 + i}`, value: `${20 + i}` })
-    await sleep(50) // sleep here is to make each xAdd get a unique, increasing timestamp
+    await sleep(50)
+  }
+
+  console.log("Loaded initial entries for all Valkey data types.")
+
+  // --- add 35,000 string keys ---
+  const TOTAL_KEYS = 35000
+  const BATCH_SIZE = 1000
```
ArgusLi: Can we move these 2 constants to the top of the file for easy reading and changing?

apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -20,3 +22,31 @@ export const calculateHotKeys = (rows) => {
   )(rows)
 }
 
+export const calculateHotKeysFromHotSlots = async (client, count = 50) => {
+  const hotSlots = await getHotSlots(client)
+
+  const slotPromises = hotSlots.map(async (slot) => {
+    const [, keys] = await client.sendCommand(["SCAN", slot["slotId"].toString(), "COUNT", "1"])
```
ArgusLi: Are we using a Sample approach? We're not getting all the hot keys in each slot as we are only getting 1 here.

apps/server/src/keys-browser.ts
ravjotbrar: Creates noise in the logs

apps/metrics/src/analyzers/calculate-hot-keys.js
ravjotbrar: Needed to adjust to keep scanning with count 1 until slot value changes in cursor. Addressed now.

apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
```diff
@@ -13,6 +13,7 @@ interface ConnectionDetails {
   alias?: string;
   role?: Role;
   clusterId?: string;
+  lfuEnabled?: boolean;
```
nassery318, line 17: just a nit - It’s not obvious why LFU matters for hot keys. A comment will help understand its significance and use.

apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -20,3 +22,46 @@ export const calculateHotKeys = (rows) => {
   )(rows)
 }
 
+export const calculateHotKeysFromHotSlots = async (client, count = 50) => {
+  const hotSlots = await getHotSlots(client)
+
+  const slotPromises = hotSlots.map(async (slot) => {
+    //console.log("Slot is: ", slot)
```
nassery318: commented console logs. might need to remove them.

apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -20,3 +22,46 @@ export const calculateHotKeys = (rows) => {
...
+      //console.log("Cursor is: ", cursor)
+      const [nextCursor, scannedKeys] = await client.sendCommand(["SCAN", cursor.toString(), "COUNT", "1"])
+      cursor = nextCursor
+      keys.push(...scannedKeys)
+      cursorToSlot = Number(cursor) & 0x3FFF
+
+    } while (cursorToSlot === slotId && cursor !== 0)
+    
+    return keys.map( async (key) => {
+      const freq = parseInt(await client.sendCommand(["OBJECT", "FREQ", key]))
```
nassery318, line 44: worth to have a comment here that OBJECT FREQ requires LFU eviction enabled in Valkey instances/cluster. Never mind, I see a comment in the PR that mentions this.

apps/metrics/src/analyzers/calculate-hot-keys.js
ravjotbrar: Removed

nassery318 (dismissed):
LGTM

PR #110, opened 2025-12-12
Add performance improvements and fix setDashboardData for clusters

ravjotbrar:
TODO: change setClusterData to only include cluster-level data

### DIFF
--- a/apps/server/src/__tests__/connection.test.ts
+++ b/apps/server/src/__tests__/connection.test.ts
@@ -95,9 +95,6 @@ describe("connectToValkey", () => {
 
       const parsedMessages = messages.map((msg) => JSON.parse(msg))
 
-      const standaloneMessage = parsedMessages[0]
-      assert.strictEqual(standaloneMessage.type, VALKEY.CONNECTION.standaloneConnectFulfilled)
-
       const clusterMessage = parsedMessages.find((msg) => msg.type === VALKEY.CLUSTER.addCluster)
       assert.ok(clusterMessage)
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/connection/ConnectionEntry.tsx
```diff
@@ -211,7 +211,7 @@ export const ConnectionEntry = ({
                 <button
                   className="flex items-center gap-1 p-2 rounded-md text-tw-primary border border-tw-primary/70
                    hover:bg-tw-primary hover:text-white"
-                  onClick={() => history.navigate(clusterId ? `/${clusterId}/${connectionId}/cluster-topology` : `/${connectionId}/dashboard`)}
+                  onClick={() => history.navigate(clusterId ? `/${clusterId}/${connectionId}/dashboard` : `/${connectionId}/dashboard`)}
                 >
                   <CircleChevronRight size={16} />
                   Open
```
nassery318, line 217: Since the entire ConnectionEntry points to a cluster, not sure we should open cluster topology from here or the dashboard. Dashboard may look consistent with the standalone instances. However, Topology will give the user an instance feedback about the cluster and perhaps they can go to the dashboard from topology. Just a thought.

ravjotbrar: I was also grappling with the same decision. This is one of those changes we can make based on feedback in the future.

nassery318 (approved):
LGTM

PR #111, opened 2025-12-12
add and update functionality for zset and stream key types

nassery318:


ravjotbrar (approved):
Only thing I ask is to update our populate script with these types so that we can reliably test them.

PR #112, opened 2025-12-15
added JSON key type viewing and deleting support

nassery318:


apps/frontend/src/components/KeyBrowser.tsx
```diff
@@ -58,6 +58,7 @@ export function KeyBrowser() {
     { value: "set", label: "Set" },
     { value: "zset", label: "Zset" },
     { value: "stream", label: "Stream" },
+    { value: "rejson-rl", label: "ReJSON-RL" },
```
ravjotbrar: I would change this to just be called json

PR #113, opened 2025-12-16
add and update JSON key type functionality

nassery318:


apps/frontend/src/components/ui/add-key.tsx
```diff
@@ -209,6 +209,13 @@ export default function AddNewKey({ onClose }: AddNewKeyProps) {
             streamEntryId: streamEntryId.trim() || undefined,
           }),
         )
+      } else if (keyType === "JSON") {
```
ravjotbrar: Let's add consts for all the key types

ravjotbrar: I would change the if/else chain into a switch case statement. I would also only set the payload in each case and dispatch once at the end.

nassery318: addressed

apps/frontend/src/components/ui/key-details-json.tsx
```diff
@@ -25,18 +36,107 @@ export default function KeyDetailsJson(
...
+
+  const handleSave = () => {
+    // validate JSON before saving
+    try {
+      JSON.parse(editedValue)
+      setError("")
+      dispatch(updateKeyRequested({
+        connectionId: connectionId,
+        key: selectedKey,
+        keyType: "JSON",
```
ravjotbrar: const

common/src/key-validators.ts
```diff
@@ -103,6 +103,28 @@ export const StreamSpec: ValidationRule[] = [
...
 ]
 
+export const JsonSpec: ValidationRule[] = [
+  ...BaseSpec,
+  {
+    validatorFn: (key) => key.keyType === "JSON" && isBlank(key.value),
+    error: "JSON value is required",
+  },
+  {
+    validatorFn: (key) => {
```
ravjotbrar, line 115: I'd rename the function or add a comment for clarity. Returning true makes me think json is valid and vice cersa

common/src/key-validators.ts
```diff
@@ -103,6 +103,28 @@ export const StreamSpec: ValidationRule[] = [
   },
 ]
 
+export const JsonSpec: ValidationRule[] = [
+  ...BaseSpec,
+  {
+    validatorFn: (key) => key.keyType === "JSON" && isBlank(key.value),
+    error: "JSON value is required",
```
ravjotbrar, line 111: const

PR #114, opened 2025-12-17
added json module detection end to end

nassery318:


apps/server/src/check-json-module.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function checkJsonModuleAvailability(
+  client: GlideClient | GlideClusterClient,
+): Promise<boolean> {
+  try {
+    const modules = await client.customCommand(["MODULE", "LIST"]) as Module[]
+
+    return modules.some((module) => {
+      const nameAttr = module.find((attr) => attr.key === "name")
+      const moduleName = nameAttr?.value
```
arseny-kostenko: can be just
```
const moduleName = module.find(attr => attr.key === "name")?.value
```
without an intermediate const

nassery318: addressed

arseny-bitquill (approved):

Issue #140, opened 2025-12-19
BUG: Cluster connection does not save on exit


ravjotbrar:
When connecting to the cluster and exiting the app, it does not save the connection for when a user reopens the app.

Issue #138, opened 2025-12-18
BUG: Unable to search responses when shape do not conform to key:value


ravjotbrar:
When a user issues a request like CONFIG GET *, the responses returned is an array of {key: keyname, value: valuename}, and our UI will only search for keys, not the value of keys. Therefore the only item searchable is the word "key" itself. 

Need to make the entire response searchable.

Issue #117, opened 2025-12-18
BUG: Hostname resolution does not work for cluster connections


ravjotbrar:
When using a hostname like localhost, instead of the associated IP address, the cluster connection results in a ConnectionNotFoundForRoute error.

PR #141, opened 2025-12-19
Fix hostname resolution, add searchable responses, and save cluster connections

ravjotbrar:
The fix for using hostnames in addition to ip addresses for clusters was to set cluster-announce-host to the host provided by the user.

The resolveHostname method is used to ensure we don't have duplicate connections for the same hostname/ipaddress pair. However, I wasn't able to test this functionality directly as localhost is not available in dns.lookup. 

I also made it so we could search values in our Send Command responses. This helps when valkey responses contain something like {key:keyname, value:valuename}.

Finally, this PR fixes cluster connections not getting saved.

Related issues:
- https://github.com/valkey-io/valkey-admin/issues/140
- https://github.com/valkey-io/valkey-admin/issues/138
- https://github.com/valkey-io/valkey-admin/issues/117

### DIFF
--- a/apps/server/src/utils.ts
+++ b/apps/server/src/utils.ts
@@ -83,9 +83,9 @@ export async function resolveHostnameOrIpAddress(hostnameOrIP: string) {
 
   const hostnameType = isIP ? "ip" : "hostname"
   try {
-    const addresses = await isIP
-      ? dns.reverse(hostnameOrIP)
-      : dns.lookup(hostnameOrIP, { family: 4, all: true }).then(R.map(R.prop("address")))
+    const addresses = isIP
+      ? await dns.reverse(hostnameOrIP)
+      : (await dns.lookup(hostnameOrIP, { family: 4, all: true })).map((result) => result.address)
 
     return { input: hostnameOrIP, hostnameType, addresses }
   } catch (err) {


apps/server/src/connection.ts
```diff
@@ -177,3 +197,20 @@ async function connectToCluster(
   )
   return clusterClient
 }
+
+async function resolveHostname(hostname: string) {
```
arseny-kostenko: rename it to `maybeHostname` or `hostnameOrIp` because this is a misleading name

apps/server/src/connection.ts
```diff
@@ -177,3 +197,20 @@ async function connectToCluster(
...
+  const isIP = /^[0-9:.]+$/.test(hostname)
```
arseny-kostenko: ~this regexp will pass for IPV6 but below you enforce `family: 4`~

apps/server/src/connection.ts
```diff
@@ -177,3 +197,20 @@ async function connectToCluster(
...
+
+  const hostnameType = isIP ? "ip" : "hostname"
+
+  try {
+    const addresses = isIP
```
arseny-kostenko: do I understand correctly that this doesn't resolve to a single invariant (like always ip addresses or always hostnames) and can produce either? because `lookup` returns the first IP address while reverse will an array of hostnames.

do we need to normalize it to a single invariant? 

given the usage of this function — does it even makes sense to return an array of `1...N` length if, effectively, it's used as "get first resolved address"

arseny-kostenko: please add a unit test for this function

apps/server/src/connection.ts
```diff
@@ -23,6 +24,24 @@ export async function connectToValkey(
     },
   ]
   try {
+    // If we've connected to the same host using IP addr or vice versa, return
+    const resolvedAddresses = (await resolveHostname(payload.host)).addresses
+    console.log("Resolved addresses: ", resolvedAddresses)
+    for (const address of resolvedAddresses) {
```
arseny-kostenko: this seems like not really a for each loop — because it short-circuits with a `return` — which makes it kinda misleading when you read the code. I.e. it's more like
```
if (resolvedAddresses.some(address => clients.has(sanitizeUrl(`${address}:${payload.port}`)))) {
  return ws.send(
    JSON.stringify({
      type: VALKEY.CONNECTION.standaloneConnectFulfilled,
      payload: { connectionId: payload.connectionId },
    }),
  )
}
// else what??? create client goes next in the code — is it expect or it should fail?
```

Based on the implementation of `resolveHostname` — the array is purely a technicality and should be a single resolved address — which eliminates the need in the loop in the first place.

apps/server/src/connection.ts
ravjotbrar: Yes that is correct. I'll change the name of the function to be more clear. I implemented this to avoid duplicate connections, given our connectionIds are based on hostname/port. So whether someone connects using an ip address or its associated hostname, we should use the same connection.

apps/server/src/connection.ts
ravjotbrar: My research here is that it's possible to have a one-to-many relationship for hostnames -> ips and the other way around. I can wrap this logic in a returnIfConnectionAlreadyExists method.

arseny-kostenko: but it doesn't iterate — it returns on the first find

apps/server/src/connection.ts
ravjotbrar: Changed

apps/server/src/utils.ts
```diff
@@ -69,3 +75,21 @@ export const parseClusterInfo = (rawInfo: ClusterResponse<string>): ParsedCluste
...
 }
+
+// Helps avoid duplicate connections
+// If user connects with IP address and then connects with hostname, we want a single connection
+export async function resolveHostnameOrIpAddress(hostnameOrIP: string) {
+  const isIP = /^[0-9:.]+$/.test(hostnameOrIP)
+
+  const hostnameType = isIP ? "ip" : "hostname"
+  try {
+    const addresses = isIP
```
arseny-kostenko: (await ..).map sneakily blends with the await dns.lookup, how about
```
const addresses = await isIP
  ? dns.reverse(hostnameOrIP)
  : dns.lookup(hostnameOrIP, { family: 4, all: true }).then(R.map(R.prop("address")))
```
so it's easier to spot that it's two promises and map is also inside the promise

apps/frontend/src/components/send-command/Response.tsx
```diff
@@ -9,8 +9,17 @@ const Response = ({ filter, response }: { filter: string, response: JSONObject }
   const filtered =
     R.pipe(
       toKeyPaths,
-      R.map((keyPath) => ({ keyPath, keyPathString: keyPath.join(".") })),
-      R.isEmpty(filter) ? R.identity : R.filter(({ keyPathString }) => keyPathString.includes(filter)),
+      R.map((keyPath) => { 
+        const value = R.path(keyPath as string[], response)
```
arseny-kostenko: don't need intermediate value
```
R.map((keyPath) => ({
  keyPath,
  keyPathString: keyPath.join("."),
  value: R.path(keyPath as string[], response),
  valueString: JSON.stringify(R.path(keyPath as string[], response)),
}))
```

apps/frontend/src/components/send-command/Response.tsx
```diff
@@ -9,8 +9,17 @@ const Response = ({ filter, response }: { filter: string, response: JSONObject }
...
+        return {
+          keyPath,
+          keyPathString: keyPath.join("."),
+          value,
+          valueString: JSON.stringify(value),
+        }
+      }),
+      // eslint-disable-next-line max-len
```
arseny-kostenko: any particular reason to suppress line length warning when it's easy to make the ternary — 3 shorter lines?

PR #142, opened 2025-12-19
Add support for Linux and WSL

allenheltondev:
Currently this app is heavily tied to Mac. I added an update to support Linux and WSL for windows.

I also added a `quickstart` file that will build and run the valkey cluster and start the frontend and server apps in one shot instead of running two separate terminals. It also auto-adds the valkey cluster as a connection so users don't have to add it manually.

Lastly, I added a brief "What is this" to the README so users know what the app is intended for.

### DIFF
--- a/apps/frontend/src/state/epics/valkeyEpics.ts
+++ b/apps/frontend/src/state/epics/valkeyEpics.ts
@@ -4,7 +4,6 @@ import * as R from "ramda"
 import { DISCONNECTED, LOCAL_STORAGE, NOT_CONNECTED, RETRY_CONFIG, retryDelay } from "@common/src/constants.ts"
 import { toast } from "sonner"
 import { sanitizeUrl } from "@common/src/url-utils"
-import type { Store } from "@reduxjs/toolkit"
 import { getSocket } from "./wsEpics"
 import {
   standaloneConnectFulfilled,
@@ -24,6 +23,7 @@ import { hotKeysRequested } from "../valkey-features/hotkeys/hotKeysSlice.ts"
 import { commandLogsRequested } from "../valkey-features/commandlogs/commandLogsSlice.ts"
 import history from "../../history.ts"
 import { setClusterData } from "../valkey-features/cluster/clusterSlice.ts"
+import type { Store } from "@reduxjs/toolkit"
 
 export const connectionEpic = (store: Store) =>
   merge(


apps/frontend/src/hooks/useAutoConnection.ts
```diff
@@ -0,0 +1,49 @@
+import { useEffect } from "react"
+import { useSelector } from "react-redux"
+import { useAppDispatch } from "./hooks"
+import { connectPending } from "@/state/valkey-features/connection/connectionSlice"
+import { sanitizeUrl } from "@common/src/url-utils"
+import { CONNECTED } from "@common/src/constants"
+import type { RootState } from "@/store"
+
+export function useAutoConnection() {
```
arseny-kostenko: Thanks for submitting a PR!

Could you please move this logic to connectionEpic at apps/frontend/src/state/epics/valkeyEpics.ts — you can `select(wsConnectFulfilled)` and map it to `connectPending` — we're trying to keep orchestration outside of components.

I can't think of a real scenario when ws connection toggles but the server->valkey connection remains in a localhost setting but still would suggest adding take(1) just in case.

README.md
```diff
@@ -1,20 +1,50 @@
 # Valkey Admin
 
-## Getting Started (Development)
+## What is Valkey Admin?
 
-### Start Valkey Cluster
+Valkey Admin is a web-based administration tool for Valkey clusters. It provides an intuitive interface to monitor, manage, and interact with your Valkey instances, offering features like real-time metrics, key management, and cluster administration.
```
ravjotbrar: We should probably remove cluster administration here, as that's not currently implemented.

README.md
```diff
@@ -1,20 +1,50 @@
...
 
-1. From the root directory, run `tools/valkey-cluster/scripts/build_run_cluster.sh`.
-    1. This will build and run the cluster.
-    2. Master hostname will be `localhost`.
-    3. Master nodes ports will be `7001` - `7003`.
+Built with React and TypeScript, Valkey Admin runs as a desktop application (via Electron) or web application, making it easy to manage your Valkey infrastructure from anywhere.
```
ravjotbrar: Some features like hotkeys and commandlogs rely on Electron, so currently we only support it as a desktop app.

ravjotbrar: Consequently, you'll need to change quickstart and the manual setup section to either remove npm run dev or clarify that it is not supported.

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -78,6 +79,39 @@ export const connectionEpic = (store: Store) =>
...
+        return null
+      }),
+      filter((action) => action !== null),
+      delay(1000),
+      tap((action) => {
+        if (action) {
+          store.dispatch(action)
+        }
+      }),
+      ignoreElements(),
```
arseny-kostenko, line 104: `tap` is to produce side-effects but here we're mapping one action to another, and for that we don't need to use dispatch (we never need to use dispatch in an epic), we simply return that another action.
```
pipe(
  select(wsConnectFulfilled),
  take(1),
  map(() => ({
     host: import.meta.env.VITE_LOCAL_VALKEY_HOST,
     port: import.meta.env.VITE_LOCAL_VALKEY_PORT,
     alias: import.meta.env.VITE_LOCAL_VALKEY_NAME || "Local Valkey Cluster",
  })), 
  filter(({ host, port }) => Boolean(host && port)),
  map(({ host, port, alias }) => connectPending({
      host,
      port: String(port),
      username: "",
      password: "",
      alias,
      connectionId: sanitizeUrl(`${host}-${port}`),
    })),
)
```

allenheltondev: Got it, thank you!

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -78,6 +79,33 @@ export const connectionEpic = (store: Store) =>
...
+            password: "",
+            alias: alias || "Local Valkey Cluster",
+            connectionId,
+          })
+        }
+
+        return null
+      }),
+      filter((action) => action !== null),
+      delay(1000)
```
arseny-kostenko: I don't we really need a manual arbitrary delay? The original comment when you had a hook was:
```
// Add a small delay to ensure WebSocket is fully ready
```
Here, we send `connectPending` only after `wsConnectFulfilled` — so we already ensure the socket is ready, and don't need a delay

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -78,6 +79,33 @@ export const connectionEpic = (store: Store) =>
...
+            username: "",
+            password: "",
+            alias: alias || "Local Valkey Cluster",
+            connectionId,
+          })
+        }
+
+        return null
+      }),
+      filter((action) => action !== null),
```
arseny-kostenko: Notice, in my previous comment it was `action -> envVarsObject -> filter -> action`;
here, you have `action -> (action | null) -> filter`

null is just a placeholder for "don't emit" — in Rx the canonical way is either filter earlier or return EMPTY.
Introducing null widens the stream type to Action | null, which weakens type safety and makes later composition more fragile.

In other words, let's keep the pipeline typing more tight, and this leaves us with two options:
1. the snippet I posted in the earlier comment (can copy paste as is, _I think_)
2. if you prefer collapsing the `-> envVarsObject -> filter ->` I had in my snippet into a single map, you still need to change map to mergeMap to return an Observable — `EMPTY | of(connectPending)`, so it becomes:

```
pipe(
  select(wsConnectFulfilled),
  take(1),
  mergeMap(() => {
    const host = import.meta.env.VITE_LOCAL_VALKEY_HOST
    const port = import.meta.env.VITE_LOCAL_VALKEY_PORT
    const alias = import.meta.env.VITE_LOCAL_VALKEY_NAME || "Local Valkey Cluster"

    if (!host || !port) return EMPTY

    return of(
      connectPending({
        host,
        port: String(port),
        username: "",
        password: "",
        alias,
        connectionId: sanitizeUrl(`${host}-${port}`),
      }),
    )
  }),
)
```

apps/frontend/src/state/epics/valkeyEpics.ts
allenheltondev: Ah yes, good catch! This was something I had added in the hook that I absentmindedly carried over. Removed.

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -78,6 +79,32 @@ export const connectionEpic = (store: Store) =>
...
+            username: "",
+            password: "",
+            alias: alias || "Local Valkey Cluster",
+            connectionId,
+          })
+        }
+
+        return null
+      }),
+      filter((action) => action !== null)
```
arseny-kostenko: one last push 😅 — let's remove the filter entirely and replace this `map` with
```
mergeMap(() => {
    const host = import.meta.env.VITE_LOCAL_VALKEY_HOST
    const port = import.meta.env.VITE_LOCAL_VALKEY_PORT
    const alias = import.meta.env.VITE_LOCAL_VALKEY_NAME || "Local Valkey Cluster"

    if (!host || !port) return EMPTY

    return of(
      connectPending({
        host,
        port: String(port),
        username: "",
        password: "",
        alias,
        connectionId: sanitizeUrl(`${host}-${port}`),
      }),
    )
  }),
  ```

File: quickstart.sh [hunk 8]
```diff
@@ -0,0 +1,138 @@
...
+    fi
+    if [ $i -eq 30 ]; then
+        echo "⚠️  Cluster health check timed out, but continuing..."
+        break
+    fi
+    echo "   Checking cluster health... ($i/30)"
+    sleep 2
+done
+
+# Step 4: Build desktop application
```
arseny-kostenko, line 14: it appears that everything above this line is the same for both quickstart.sh and quickstart-web.sh — can we extract it maybe?

apps/frontend/src/state/epics/valkeyEpics.ts
allenheltondev: I tried implementing this with mergeMap + of() + EMPTY, but found that this codebase uses custom epic middleware rather than standard redux-observable.

The mergeMap approach creates and returns actions correctly, but they don't get dispatched to the store because the custom middleware doesn't automatically dispatch returned actions like redux-observable does.

Looking at the other epics in the codebase, they all use the tap + store.dispatch() + ignoreElements() pattern, so I've kept that approach for consistency. The delay has been removed as you suggested since the WebSocket connection is guaranteed to be ready when wsConnectFulfilled fires.

arseny-kostenko: ah, you're right, my bad — for some reason I was thinking in redux-observable terms; but IIRC we didn't go with it because it doesn't seem to be very active, and decided to not have a possibly soon-obsolete dependency

arseny-bitquill (dismissed):
LGTM

@ravjotbrar, can you take a look as well?

ravjotbrar (approved):

arseny-bitquill (dismissed):

allenheltondev:
I finally got eslint configured properly. All linting issues are resolved. Thank you @arseny-bitquill

ravjotbrar (approved):

Issue #66, opened 2025-11-07
FEATURE: Implement ability to edit connection


ravjotbrar:
Edit connection should open up the same connection dialogue when clicking 'Add a Connection', except with the non-sensitive fields already pre-populated. If core connection details change, it will create a new connection in the backend with new values.

PR #154, opened 2025-12-29
#66 - Implement ability to edit connections

allenheltondev:
# Summary
Fix connection editing so connections don't vanish when you changing config. And to unify the experience between adding and editing connections

## Changes
* Replace inline editing with proper edit modal
* Preserve connection history when editing host/port/username/password
* Add preservedHistory param to connectPending action
* Fix undefined connectionDetails crash in standaloneConnectFulfilled
* Fix jsonModuleAvailable being set to wrong value (lfuEnabled instead of jsonModuleAvailable)
* Clean up ConnectionEntry component - remove inline editing complexity
* Add silent delete option for seamless connection replacement during edits

## Why
Implementing issue #66  

Editing a connection would delete the old one and create a new one with no history. If the new config was bad, the connection disappeared from the UI since we only show connections with history. Even worse, if you fixed the config, it would still fail due to a WebSocket error from undefined payload handling.

Now edited connections keep their history and stay visible even when broken, so you can actually fix them.


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A modern admin interface for Valkey databases",
   "private": true,
   "version": "0.0.0",
+  "type": "module",
```
ravjotbrar: Unfortunately, this breaks notarizaton, because it still relies on CommonJS.

PR #156, opened 2025-12-30
Expose config options and add an updateConfig endpoint

ravjotbrar:
#### Goal
Before this change, users had no way to enable or disable monitoring for hot key collection. This PR aimed to expose the relevant config params to allow users control over how long they want monitor to sample for, if it at all.

#### Details
This change adds a config slice that keeps state of client, valkey, and metrics server config properties. In addition, it adds an updateConfig handler to our backend that sends an api call to the newly created update-config endpoint in the metrics server. 

For now, we only expose enableMonitor and monitorDuration to control hotkey collection using the monitor approach. By default, monitor is disabled, as it is considered an expensive command. If users do not have monitor enabled and they don't have the correct settings set for our hot slots approach for calculating hotkeys, we let them know by displaying a message.

If a user changes a setting when connected to a cluster, that setting is changed for all nodes. There is also an action that enabled cluster slot stats on all nodes, however, the design for how we expose that to users is not confirmed. For now, users can set a config property using Send Command in the app and it will automatically redirect to all nodes in a cluster

### TODO
I avoided creating more tests in this PR simply to keep the size of this manageable. I will create another PR adding tests for all the handlers in our backend.

### DIFF
--- a/apps/frontend/src/state/valkey-features/config/configSlice.ts
+++ b/apps/frontend/src/state/valkey-features/config/configSlice.ts
@@ -12,15 +12,17 @@ interface MonitorConfig {
   monitorEnabled: boolean, 
   // How long to monitor before stopping (ms)
   monitorDuration: number,
+
+  // TODO: expose monitorInterval and continuousMonitoring
   // How long to wait before monitoring again when using continuous mode (ms)
-  //monitorInterval: number,
+  monitorInterval?: number,
   // Default is one cycle and then turn off monitoring
-  //continuousMonitoring: boolean,
+  continuousMonitoring?: boolean,
 }
 interface ConfigState {
   [connectionId: string]: {
     darkMode: boolean,
-    // Valkey related. Won't expose yet.
+    // Valkey related. TODO: find best way to expose to user
     keyEvictionPolicy?: KeyEvictionPolicy
     clusterSlotStatsEnabled?: boolean,
     pollingInterval: number, 
@@ -46,8 +48,6 @@ const configSlice = createSlice({
           monitoring: {
             monitorEnabled: false, 
             monitorDuration: 6000,
-          //monitorInterval: 20000, 
-          //continuousMonitoring: false,
           },
           status: "updated",
         }


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -303,3 +305,20 @@ export const getCommandLogsEpic = (store: Store) =>
     }),
     ignoreElements(),
   )
+
+export const updateConfigEpic = (store: Store) => 
+  action$.pipe(
+    select(updateConfig),
+    tap((action) => {
```
arseny-kostenko, line 357: leave a comment what's happening here

apps/metrics/src/config.js
```diff
@@ -41,5 +42,66 @@ const loadConfig = () => {
 
   return cfg
 }
+const getConfig = () => config ? config : loadConfig() 
+
+const updateConfig = (partialConfig) => {
+  try {
+    validatePartialConfig(partialConfig)
```
arseny-kostenko: this is an implicit side-effect with an error. We should be explicit with errors. While we're not adding a monadic return type like https://effect.website/docs/getting-started/the-effect-type/ (read this), we still want errors to be explicit via return types; make validatePartialConfig return an Error or null, so it's explicit and the user doesn't have to inspect how validatePartialConfig works to guess that updateConfig may actually throw

arseny-kostenko: especially, that you're catching an Error and return the same object anyway — there is no need to throw and catch, if you can instanceof Error like you already do in the catch

apps/metrics/src/config.js
```diff
@@ -41,5 +42,66 @@ const loadConfig = () => {
...
+    console.error("Failed to update config:", error.message)
+    return {
+      success: false,
+      message: error instanceof Error ? error.message : String(error),
+      data: error,
+    }
+  }
+}
+
+const validatePartialConfig = (partialConfig) => {
```
arseny-kostenko, line 69: I feel like we need to start using [zod](https://zod.dev/) instead of building our own bycicles

apps/frontend/src/components/ui/hot-keys.tsx
```diff
@@ -160,15 +161,16 @@ export function HotKeys({ data, status, onKeyClick, selectedKey }: HotKeysProps)
...
-          <span className="text-sm mt-1">
-            <span className="font-semibold">Note:</span>
-            {" "}To collect hotkeys, you must either have
-            <span className="font-semibold"> monitoring enabled</span>
-            {" "}in Settings or use an
-            <span className="font-semibold"> LFU eviction policy</span>
-            {" "}with
-            <span className="font-semibold"> cluster slot stats enabled</span>.
-          </span>
+          {errorMessage && (
```
nassery318, line 164: Not sure about this: in Line 162 we say No Hot Keys Found and in case of error both the Note (now moved to server side) and No Hot keys Found will apear at the same time.

apps/frontend/src/state/valkey-features/config/configSlice.ts
```diff
@@ -0,0 +1,75 @@
...
+  R.path([VALKEY.CONFIG.name, id], state)
+
+interface MonitorConfig {
+  monitorEnabled: boolean, 
+  // How long to monitor before stopping (ms)
+  monitorDuration: number,
+  // How long to wait before monitoring again when using continuous mode (ms)
+  //monitorInterval: number,
+  // Default is one cycle and then turn off monitoring
+  //continuousMonitoring: boolean,
```
nassery318: We can clean up or mark them as TO DO

apps/frontend/src/state/valkey-features/config/configSlice.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (!state[connectionId]) {
+        state[connectionId] = {
+          darkMode: false, 
+          keyEvictionPolicy, 
+          clusterSlotStatsEnabled: clusterSlotStatsEnabled ?? false, 
+          pollingInterval: 5000,
+          monitoring: {
+            monitorEnabled: false, 
+            monitorDuration: 6000,
+          //monitorInterval: 20000,
```
nassery318: Same here

apps/frontend/src/state/valkey-features/config/configSlice.ts
ravjotbrar: Good point.

apps/frontend/src/components/ui/hot-keys.tsx
ravjotbrar: Will change depending on user feedback.

Issue #128, opened 2025-12-18
UX: Missing refresh button for HotKeys


ravjotbrar:
Add refresh button for hotkeys similar to commandlogs. If we are unable to, we should provide some other feedback to the user.

PR #157, opened 2025-12-31
feature(monitoring): add refresh button for hot keys tab

KIvanow:
## Summary
Adds a refresh button to the Hot Keys tab in the Monitoring page, matching the UX of the Command Logs tab.

## Changes
- Added `refreshHotKeys()` function that dispatches `hotKeysRequested` action
- Added refresh button UI that appears when Hot Keys tab is active
- Uses same styling as Command Logs refresh button for consistency

## Testing
1. Navigate to Monitoring page
2. Select Hot Keys tab
3. Click Refresh button
4. Verify monitoring cycle starts and hot keys are updated after ~10 seconds

Ref #128

### DIFF
--- a/apps/frontend/src/state/valkey-features/hotkeys/hotKeysSlice.ts
+++ b/apps/frontend/src/state/valkey-features/hotkeys/hotKeysSlice.ts
@@ -34,11 +34,13 @@ const hotKeysSlice = createSlice({
       if (!state[connectionId]) {
         state[connectionId] = {
           hotKeys: [],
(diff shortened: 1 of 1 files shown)


KIvanow:
> Thanks for the PR! When you click the refresh button for hotkeys, ideally we'd like it to show the loading state, similar to how it does the first time it's fetching hotkeys.
> 
> If you're having trouble building the desktop app, I would rebase onto main to get the latest fixes.

Added it. Let me know what you think

Issue #127, opened 2025-12-18
UX: Hover state unnecessary on Dashboard metrics


ravjotbrar:
If the dashboard metrics are not clickable or if there is no hover functionality, then this can be removed.

PR #158, opened 2025-12-31
fix(dashboard): remove unnecessary hover state from metrics

KIvanow:
Dashboard metrics are not clickable and have no hover functionality, so the hover effect was misleading to users. The Accordion component is used nowhere else in the project, so there is no chance of regression.

Ref #127


nassery318 (commented):
Good point. Since the aim is to help the eye track across from label to value in case the user is looking into accordions with more metrics (Client Connectivity Metrics), perhaps a subtle divider (border-b) could help.

KIvanow:
> Good point. Since the aim is to help the eye track across from label to value in case the user is looking into accordions with more metrics (Client Connectivity Metrics), perhaps a subtle divider (border-b) could help.

Good suggestion @nassery318 here is how it looks like 
image

PR #160, opened 2026-01-01
UX: Reorder command log columns to show command first

### LINKED ISSUE #129: UX: Make command first column in Command Logs
At the moment, the first column is timestamp. We should organize columns in order of importance:  command --> duration --> timestamp etc...

KIvanow:
## Summary
- Reorders columns in Command Logs table from `Timestamp --> Duration --> Command --> Client` to `Command --> Duration --> Timestamp --> Client`
- Prioritizes the most important information (the command) as the first column
- Centers the Timestamp column text to align with Duration and Client Address columns

Fixes #129


nassery318 (commented):
Only linting issues, other than that looks good to me.

PR #162, opened 2026-01-03
Added `tolerance` parameter for cpuReduce so that similar values are collapsed

arseny-kostenko:
### Summary

Adds tolerance-based point suppression to CPU usage calculation and exposes it via the `/cpu` API to reduce visual noise in charts (and/or take move this calculation from UI to the backend).

### Details

- Extends the CPU reducer to optionally skip points within a configurable tolerance of the previous value
- Preserves transitions to/from zero (only skips repeated zero values)
- Exposes `tolerance` as a query parameter on the `/cpu` endpoint
- Applies sane defaults and bounds (defaults to 2.5%, capped at 20%)
- Updates tests to cover tolerance behavior

### DIFF
--- a/apps/metrics/src/index.js
+++ b/apps/metrics/src/index.js
@@ -47,11 +47,11 @@ async function main() {
       const tolerance = R.pipe(
         R.pathOr("0.025", ["query", "tolerance"]),
         Number,
+        Math.abs, // no negative numbers
         R.when( // when not a number or more than 20% — default to 2.5% tolerance interval
           R.either(Number.isNaN, R.lte(0.2)),
           R.always(0.025),
         ),
-        Math.abs, // no negative numbers
       )(_req)
 
       const series = await Streamer.info_cpu(cpuFold({ tolerance }))


apps/metrics/src/analyzers/calculate-cpu-usage.js
```diff
@@ -17,64 +17,90 @@ export const cpuSeed = {
 
 export const cpuFilter = ({ metric }) => metric === "used_cpu_sys" || metric === "used_cpu_user"
 
+const shouldEmitPoint = ({ prevValue, currValue, tolerance = 0 }) =>
+  prevValue == null || // first point
+  (prevValue === 0 && currValue !== 0) || // went from zero to something
+  (prevValue !== 0 && currValue === 0) || // went from something to zero
+  (Math.abs((currValue - prevValue) / prevValue) > tolerance) // "noticeably" changed
```
ravjotbrar, line 24: Appreciate the comments.

apps/metrics/src/index.js
```diff
@@ -43,12 +44,17 @@ async function main() {
...
-        filterFn: cpuFilter,
-        reducer: cpuReducer,
-        seed: cpuSeed,
-        finalize: cpuFinalize,
-      })
+      const tolerance = R.pipe(
+        R.pathOr("0.025", ["query", "tolerance"]),
+        Number,
+        R.when( // when not a number or more than 20% — default to 2.5% tolerance interval
+          R.either(Number.isNaN, R.lte(0.2)),
```
ravjotbrar, line 52: What if a user passes a negative number here? It's a number and it's less than 0.2, so theoretically the tolerance could go higher than 20%.

arseny-kostenko: addressed by moving Math.abs up in the pipe

Issue #139, opened 2025-12-18
BUG: Diff command doesn't work for nested keys


ravjotbrar:
If we issue an info command using the cluster client, it'll parse the keys as the node address and the value as the entire info object for that node. We need to parse the keys of the info object per node.

PR #171, opened 2026-01-08
Fixed parsing of fanout response

arseny-kostenko:
## Description

Fixed [defect](https://github.com/valkey-io/valkey-admin/issues/139) where a diff of cluster commands would not work properly because cluster commands can return with a fanout shape, i.e. per node, so the actual content per node remained unparsed as big flat strings. I updated the server response logic that detects that and parses the content.

### Change Visualization

N/A

Before 
Screenshot 2026-01-08 at 3 11 07 PM

After
Screenshot 2026-01-08 at 3 08 39 PM

### DIFF
--- a/apps/server/src/send-command.ts
+++ b/apps/server/src/send-command.ts
@@ -14,29 +14,23 @@ export async function sendValkeyRunCommand(
 ) {
   try {
     const response = await client.customCommand(payload.command.split(" "))
-
-    if (isRequestError(response)) {
-      ws.send(
-        JSON.stringify({
-          meta: { command: payload.command },
-          type: VALKEY.COMMAND.sendFailed,
-          payload: response,
-        }),
-      )
-    }
-
-    const parsed = parseResponse(response)
+    const isError = isRequestError(response)
 
     ws.send(
       JSON.stringify({
         meta: {
-          connectionId: payload.connectionId,
           command: payload.command,
+          connectionId: payload.connectionId,
         },
-        type: VALKEY.COMMAND.sendFulfilled,
-        payload: parsed,
+        type: isError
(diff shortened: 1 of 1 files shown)


apps/server/src/send-command.ts
```diff
@@ -3,36 +3,38 @@ import WebSocket from "ws"
...
-      response = parseResponse(response)
+    const response = await client.customCommand(payload.command.split(" "))
+
+    if (isRequestError(response)) {
+      ws.send(
+        JSON.stringify({
+          meta: { command: payload.command },
+          type: VALKEY.COMMAND.sendFailed,
+          payload: response,
+        }),
```
ravjotbrar: Should we return here? Seems like we send a failed a fulfilled message if theres a responserror.

apps/server/src/actions/cpuUsage.ts
```diff
@@ -43,9 +43,14 @@ const sendCpuUsageError = (
...
 
+type RequestPayload = {
+  connectionIds: string[]
+  timeRange?: string
+}
+
 export const cpuUsageRequested = withDeps<Deps, void>(
   async ({ ws, metricsServerURIs, action }) => {
-    const { connectionIds, timeRange = "12h" } = action.payload
+    const { connectionIds, timeRange = "12h" } = action.payload as unknown as RequestPayload
```
ravjotbrar, line 53: Is this to satisfy the compiler?

arseny-kostenko: yes, it was red in Nas's PR, I just fixed it

apps/server/src/send-command.ts
arseny-kostenko: good point

PR #172, opened 2026-01-09
Create frontend unit tests

ArgusLi:
## Description

Add frontend unit tests.
Add frontend tests to test workflow.

### Change Visualization

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,3 +22,5 @@ jobs:
       run: npm ci
     - name: Test Server
       run: npm test --workspace=apps/server
+    - name: Test Frontend
+      run: npm test --workspace=apps/frontend
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -7,7 +7,7 @@
     "build": "vite build",
     "preview": "vite preview",
     "lint": "eslint .",
-    "test": "vitest",
+    "test": "vitest run",
     "test:ui": "vitest --ui",
     "test:run": "vitest run",
     "test:coverage": "vitest run --coverage",


apps/frontend/src/hooks/useIsConnected.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    })
+
+    const { result } = renderHook(() => useIsConnected(), {
+      wrapper: ({ children }) => <Provider store={store}>{children}</Provider>,
+    })
+
+    expect(result.current).toBe(true)
+  })
+
+  it("should return true when status is DISCONNECTED", () => {
```
ravjotbrar, line 45: Hmm maybe we should rename the hook. Sounds confusing if it always returns true no matter the status.

ArgusLi: I think we should keep this for now. Once we add disconnection logic, then it'll make sense.

apps/frontend/src/state/valkey-features/connection/connectionSlice.test.ts
```diff
@@ -0,0 +1,632 @@
+import { describe, it, expect, beforeEach } from "vitest"
```
ravjotbrar, line 1: Are we intentionally only testing a select few slices?

apps/frontend/src/hooks/useIsConnected.test.tsx
ravjotbrar: I thought we already had disconnect logic

ArgusLi: Last I checked, it just marks it as disconnected, but we maintain the connection.

apps/frontend/src/state/valkey-features/connection/connectionSlice.test.ts
ArgusLi: These are just the Slices that seem to be the most important.

arseny-kostenko (approved):
lgtm

Issue #118, opened 2025-12-18
Feature: Add sort capability to Key Browser


ravjotbrar:
Add ability to sort by name, ttl, or size for list of keys sampled in key browser

PR #176, opened 2026-01-12
FEAT: Add sorting to key browser

allenheltondev:
## Description

Adding sorting by name, ttl, or size to the key browser. Click once for sorting descending, click again for ascending. This is work for #118.

Added the first unit tests in the common directory, so I updated the GH workflow to include that workspace during the test step.

### Change Visualization

Screenshots of ascending and descending name sort 👇 
image
image


apps/frontend/src/components/ui/key-sort-control.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  return (
+    <div className="flex items-center gap-1">
+      {sortOptions.map(({ field, label, icon }) => {
+        const isActive = currentSort.field === field
+        const direction = isActive ? currentSort.direction : "asc"
+
+        return (
+          <Button
```
nassery318, line 49: I would make the buttons look consistent with other items. 
image

apps/frontend/src/components/ui/key-sort-control.tsx
```diff
@@ -0,0 +1,71 @@
...
+            className="flex items-center gap-1.5"
```
nassery318, line 50: Maybe a bit polished but something like:

`<button
            className={`flex items-center gap-1 h-10 mr-1 px-4 py-2 dark:border-tw-dark-border border rounded ${isActive ? "bg-tw-primary/70 text-white" : "bg-white dark:bg-gray-800"}`}
            disabled={disabled}
            key={field}
            onClick={() => handleSortClick(field)}
          >
            {icon}
            {label}
            {isActive && (
              direction === "asc" ? (
                <ArrowUp size={12} />
              ) : (
                <ArrowDown size={12} />
              )
            )}
          </button>`

common/src/key-sorting.ts
```diff
@@ -0,0 +1,95 @@
...
+  direction: SortDirection
+}
+
+/**
+ * Sorts an array of keys based on the specified sort option
+ * @param keys Array of key information objects
+ * @param sortOption Sort configuration specifying field and direction
+ * @returns New sorted array of keys
+ */
+export function sortKeys(keys: KeyInfo[], sortOption: SortOption): KeyInfo[] {
```
arseny-kostenko, line 23: we can reduce boilerplate by leveraging Ramda's OOB functions 
```
import * as R from "ramda"

export function sortKeys(
  keys: KeyInfo[],
  sortOption: SortOption
): KeyInfo[] {
  if (!keys || keys.length === 0) {
    return keys
  }

  const direction = sortOption.direction === "desc" ? R.descend : R.ascend

  const sorters = {
    name: direction<KeyInfo>(R.pipe(R.prop("name"), R.toLower)),

    ttl: direction<KeyInfo>(
      R.pipe(
        R.propOr(-1, "ttl"),
        (ttl: number) => (ttl === -1 ? Number.POSITIVE_INFINITY : ttl)
      )
    ),

    size: direction<KeyInfo>(R.propOr(0, "size")),
   }

  const sorter = sorters[sortOption.field] ?? R.ascend<KeyInfo>(() => 0)

  return R.sortWith([sorter], keys)
}

```

PR #177, opened 2026-01-13
Memory endpoint for timeseries

arseny-kostenko:
## Description
This is the Metrics service changes.

Added support of:
1. `since`, `until` params for /cpu and /memory endpoints to limit timeseries output
2. added `maxPoints` param for downsampling
3. Implemented minmax downsampling
4. Removed downsampling and time limiting from the bff server
5. Added http client file
6. Added zod for validation

### Change Visualization

N/A API change

### DIFF
--- a/apps/metrics/src/utils/helpers.js
+++ b/apps/metrics/src/utils/helpers.js
@@ -60,10 +60,18 @@ export const downsampleMinMaxOrdered = R.curry(({ maxPoints }, series) => {
   }
 
   const out = buckets.reduce((acc, b) => {
-    if (b.min) acc.push(b.min)
-    if (b.max) acc.push(b.max)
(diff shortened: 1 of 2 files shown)


apps/metrics/src/analyzers/memory-metrics.js
```diff
@@ -0,0 +1,112 @@
+import * as R from "ramda"
+import { downsampleMinMaxOrdered } from "../utils/helpers.js"
+
+export const MEMORY_METRICS = {
+  // core usage
+  total_allocated: {
+    key: "allocated_bytes",
+    description: "How much memory Valkey is actually using for data and internal structures.",
```
ravjotbrar, line 8: big fan of these easily readable descriptions

PR #179, opened 2026-01-13
Add support for configuring TLS connections

ravjotbrar:
## Description

This change allows users to use TLS when connecting to their Valkey server. There is also an option to enable/disable certificate verification if they are using a self-signed cert. 

In order to use the same connection properties in the metrics server, I had to change the valkey client to the same GlideClient we use in our application. 

Additionally, I refactored the way we store our connection details so that future modifications need to touch less code.

### DIFF
--- a/apps/metrics/src/index.js
+++ b/apps/metrics/src/index.js
@@ -1,6 +1,5 @@
 import fs from "node:fs"
 import express from "express"
-import * as R from "ramda"
 import { getConfig, updateConfig } from "./config.js"
 import * as Streamer from "./effects/ndjson-streamer.js"
 import { setupCollectors } from "./init-collectors.js"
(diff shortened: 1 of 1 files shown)


apps/server/src/index.ts
```diff
@@ -66,25 +87,6 @@ wss.on("connection", (ws: WebSocket) => {
       console.log("Failed to parse the message", message.toString(), e)
     }
 
-    const handlers: Record<string, Handler> = {
```
ravjotbrar, line 69: I moved this out of ws.on message to avoid recreating handlers every time a message is sent.

apps/metrics/src/index.js
```diff
@@ -103,7 +122,7 @@ async function main() {
   const server = app.listen(port, () => {
     const assignedPort = server.address().port
     console.log(`listening on http://0.0.0.0:${assignedPort}`)
-    process.send?.({ type: "metrics-started", payload: { valkeyUrl: url, metricsHost: "http://0.0.0.0", metricsPort: assignedPort } })
+    process.send?.({ type: "metrics-started", payload: { metricsHost: "http://0.0.0.0", metricsPort: assignedPort } })
```
ravjotbrar, line 120: We should probably add an env variable to configure the metrics host

arseny-kostenko (approved):

PR #180, opened 2026-01-13
UX:replaced TBD measures with actual metrics

nassery318:
## Description

1. Previously Operations and Hit Ratio were TBD items in KeyBrowser 
2. Now replaced by actual measures from the info command
3. Added tooltips and descriptions for better UX

addresses the issue - REF: #175 

### Change Visualization

image


arseny-kostenko (dismissed):

common/src/cache-hit-ratio.ts
arseny-kostenko, line 1: can you rename it to kebab case?

nassery318: changed to kebab case

arseny-kostenko (approved):

Issue #122, opened 2025-12-18
UX: remove connect/disconnect (power) button


ravjotbrar:
In the connection management page, once you've connected to a node you have the option to disconnect. At the moment it's unclear what value this provides and users may assume this is used to turn nodes on and off.

Issue #120, opened 2025-12-18
UX: Cluster name should be a hyperlink similar to standalone connection names


ravjotbrar:
At the moment, the user has to click the cluster dropdown and choose which connection they want to connect to. Clicking this hyperlink should bring you into either the cluster topology page, or pick a node part of the cluster and route to its dashboard.

Issue #119, opened 2025-12-18
UX: Topology dropdown should not wrap connection name


ravjotbrar:
Dropdown could dynamically adjust to accomadate longer names or we could truncate with ellipses and have a hover over feature that displays full connection name.

PR #181, opened 2026-01-13
Minor UX updates

allenheltondev:
## Description

Updates to address three small UX issues in the backlog.

* #119 - Don't wrap connection name in topology dropdown, truncate it with ellipses and display a tooltip with the full name on hover
* #120 - Cluster name in connection view is now clickable and brings you to topology view (with explainer tooltip)
* #122 - Change connect/disconnect icons to be more intuitive and added text labels

Issue 122 recommends deleting the functionality, but it is useful. I just updated it to be more intuitive.

### Change Visualization

New disconnect button instead of an on/off button
image
Truncated name + link to topology
image


allenheltondev:
This has been approved for a while, could someone please merge it for me? @nassery318

Issue #151, opened 2025-12-25
FEATURE: Command Auto-Suggestions


sandeepkunusoth:
Add auto complete for Valkey commands with Command suggestions as we type

PR #184, opened 2026-01-15
Feature: autocomplete commands

allenheltondev:
## Description

Adding an autocomplete/intellisense for the command page. Works for commands only (no resources), but does tell you the parameters required for the command. This is for issue #151 

### Change Visualization

Command page with autocomplete open
image

### DIFF
--- a/apps/frontend/src/components/key-browser/KeyBrowser.tsx
+++ b/apps/frontend/src/components/key-browser/KeyBrowser.tsx
@@ -155,7 +155,7 @@ export function KeyBrowser() {
             className="flex-1"
             label="Total Keys"
             tooltip={
-              <TooltipIcon description="Total number of keys in the database" size={14} />
+              <TooltipIcon description="Total number of keys sampled" size={14} />
             }
             value={totalKeys}
           />
@@ -163,7 +163,7 @@ export function KeyBrowser() {
             className="flex-1"
             label="Memory Usage"
             tooltip={
-              <TooltipIcon description="Memory used by all keys in the database" size={14} />
+              <TooltipIcon description="Memory used by keys sampled" size={14} />
             }
             value={formatBytes(totalMemoryUsage)}
           />
@@ -206,7 +206,7 @@ export function KeyBrowser() {
             disabled={loading}
             onChange={(e) => setSearchPattern(e.target.value)}
             onClear={handleClearSearch}
-            placeholder="Search keys (use * to search patterns like user:*)"
+            placeholder="Search keys (use * to search patterns like user:* and press Enter)"
             value={searchPattern}
           />
         </form>
--- a/apps/frontend/src/components/key-browser/key-details/key-details.tsx
+++ b/apps/frontend/src/components/key-browser/key-details/key-details.tsx
@@ -1,4 +1,4 @@
-import { Key, Trash, X } from "lucide-react"
+import { Key, Trash, TriangleAlert, X } from "lucide-react"
 import { useState } from "react"
 import { convertTTL } from "@common/src/ttl-conversion"
 import { formatBytes } from "@common/src/bytes-conversion"
@@ -17,12 +17,14 @@ import { useAppDispatch } from "@/hooks/hooks"
 import { deleteKeyRequested } from "@/state/valkey-features/keys/keyBrowserSlice"
 import { CustomTooltip } from "@/components/ui/tooltip"
 import { Typography } from "@/components/ui/typography"
+import { Alert, AlertDescription } from "@/components/ui/alert"
 
 interface BaseKeyInfo {
   name: string;
   ttl: number;
   size: number;
   collectionSize?: number;
+  elementsWarning?: string;
 }
 
 interface ElementInfo {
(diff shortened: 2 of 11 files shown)


apps/frontend/src/components/ui/autocomplete-dropdown.tsx
```diff
@@ -0,0 +1,235 @@
...
+  isLoading: boolean;
+  onSelect: (command: ValkeyCommand) => void;
+  onClose: () => void;
+  inputRef: React.RefObject<HTMLTextAreaElement>;
+}
+
+interface CommandSuggestionProps {
+  matchResult: MatchResult;
+  isSelected: boolean;
+  query: string;
```
arseny-kostenko: this doesn't seem to be used

apps/frontend/src/components/ui/autocomplete-dropdown.tsx
```diff
@@ -0,0 +1,235 @@
...
+          <div
+            key={matchResult.command.name}
+            ref={index === selectedIndex ? selectedItemRef : null}
+          >
+            <CommandSuggestion
+              isSelected={index === selectedIndex}
+              matchResult={matchResult}
+              onClick={() => onSelect(matchResult.command)}
+              optionId={`${dropdownId}-option-${index}`}
+              query=""
```
arseny-kostenko: unnecessary?

apps/frontend/src/components/ui/autocomplete-dropdown.tsx
allenheltondev: Good catch. that's dead code - i went with a different way for highlighting matched strings

apps/frontend/src/hooks/useValkeyAutocomplete.ts
```diff
@@ -0,0 +1,243 @@
...
+  }, [suggestions.length])
+
+  const insertCommand = useCallback((command: ValkeyCommand, inputRef: React.RefObject<HTMLTextAreaElement>) => {
+    if (!inputRef.current) return
+
+    const textarea = inputRef.current
+    const currentValue = textarea.value
+    const cursorPosition = textarea.selectionStart || 0
+
+    const lineStart = currentValue.lastIndexOf("\n", cursorPosition - 1) + 1
```
arseny-kostenko: this probably can be inlined on line 93

apps/frontend/src/components/send-command/SendCommand.tsx
```diff
@@ -28,13 +30,131 @@ export function SendCommand() {
...
+          if (autocomplete.state.suggestions.length > 0) {
+            e.preventDefault()
+            const selectedCommand = autocomplete.state.suggestions[autocomplete.state.selectedIndex]?.command
+            if (selectedCommand) {
+              insertSelectedCommand(selectedCommand)
+              return
+            }
+          }
+          break
+        case "Tab":
```
arseny-kostenko: looks like Tab and Enter can be collapsed since their handlers are identical
case "Enter":
case "Tab": {

apps/frontend/src/hooks/useValkeyAutocomplete.ts
```diff
@@ -0,0 +1,243 @@
...
+      commandText += ` ${placeholders}`
+    } else if (existingArgs.trim()) {
+      commandText += ` ${existingArgs}`
+    }
+
+    const beforeCommand = currentValue.substring(0, commandStart)
+    const afterLine = currentValue.substring(lineEnd)
+    const newValue = beforeCommand + commandText + afterLine
+
+    textarea.value = newValue
```
arseny-kostenko: from this point, this feels very off tbh — we have a react-state-managed textarea component and suddenly it's force-fed directly through DOM, and an event impersonating a user, while we have setState.

Essentially, we have a split-brain between the hook and component states

apps/frontend/src/hooks/useValkeyAutocomplete.ts
```diff
@@ -0,0 +1,243 @@
...
+    insertCommand,
+    hide,
+    show,
+    navigateUp,
+    navigateDown,
+    navigateToFirst,
+    navigateToLast,
+  }
+
+  return { state, actions }
```
arseny-kostenko: 1. This hook in general does way too many things in a intertwined way. Also, I cannot see how this hook can be reused by other components that don't do "send command", so a hook seems like a wrong abstraction. 

2. Given that we're building a specialized component that does autocompletion of valkey commands, it seems that the right abstraction is still a component wrapping the textarea where the user types, the autocompletion dropdown, and contains the logic and the shared state (textarea.value, cursor position, commands matches)

3. Given the reactive nature of this flow, you had to add debouncing via setTimeouts and all that which makes the logic hard to follow, so I suggest leveraging the fact that we already have rxjs where these operators are OOB. You can create a subject and feed in onChange of textarea and pipe it to debounce, map with all the transforms and filter and in subscribe do setSuggestions

arseny-kostenko:
Thanks @allenheltondev, this looks way better! I'll do a more thorough review next week. Meanwhile, could you leave some comments around text insertion util — how it works and etc. Thank you!

apps/frontend/src/components/ui/autocomplete-dropdown.tsx
```diff
@@ -0,0 +1,185 @@
...
+    >
+      {isLoading ? (
+        <div
+          aria-live="polite"
+          className="px-3 py-2 text-sm text-muted-foreground"
+          role="status"
+        >
+          Loading suggestions...
+        </div>
+      ) : suggestions.length === 0 ? (
```
arseny-kostenko: let's collapse loading suggestions and no commands found given that it's the same compont
```
const statusMessage = isLoading
  ? "Loading suggestions..."
  : suggestions.length === 0
    ? "No commands found"
    : null

...

{
   statusMessage ?
  <div aria-live="polite" ....>
    {statusMessage}
  </div> :
  suggestions.map((matchResult, index) => ....)
}

```

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
+import type { ValkeyCommand, MatchResult } from "@/types/valkey-commands"
+import valkeyCommands from "@/data/valkey-commands.json"
+
+/**
+ * Calculate Levenshtein distance between two strings
+ */
+function levenshteinDistance(str1: string, str2: string): number {
```
arseny-kostenko: let's switch to https://www.npmjs.com/package/fastest-levenshtein

this current Levenshtein implementation builds a full (n × m) matrix for every comparison, which is O(n x m) in both time and memory. Since this runs on each keystroke across multiple candidates, it creates unnecessary allocations and GC pressure.

fastest-levenshtein uses a Myers bit-parallel algorithm that avoids building the full matrix, so it's much faster on small and medium strings and has lower memory overhead, and is well-tested

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  if (matchType === "contains") {
+    const index = lowerText.indexOf(lowerQuery)
+    if (index !== -1) {
+      return [[index, index + query.length]]
+    }
+  }
+
+  // For fuzzy matches, we'll highlight the entire command name for simplicity
+  if (matchType === "fuzzy") {
```
arseny-kostenko: since we're filtering matched commands anyway — highlighting the entire fuzzy matched command probably it doesn't help? it's like all of them will be highlighted. Consider returning []?

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+      matchType = "prefix"
+      score = lowerQuery.length / lowerCommandName.length // Higher score for longer matches
+    }
+    // Contains match
+    else if (lowerCommandName.includes(lowerQuery)) {
+      matchType = "contains"
+      score = 0.5 + (lowerQuery.length / lowerCommandName.length) * 0.3
+    }
+    // Fuzzy match using Levenshtein distance
+    else {
```
arseny-kostenko: maybe worth allowing fuzzy matching only if the query is at least lowerQuery.length >= 3?

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  const trimmedQuery = query.trim()
+  const lowerQuery = trimmedQuery.toLowerCase()
+  const results: MatchResult[] = []
+
+  // Cast the imported JSON to the correct type
+  const commands = getCommands({ adminMode })
+
+  for (const command of commands) {
+    const lowerCommandName = command.name.toLowerCase()
```
arseny-kostenko: given that in the valkey-commands.json all commands are uppercase, maybe it's worth switching to upperQuery so that we don't have to switch casing for command.name?

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+      score = lowerQuery.length / lowerCommandName.length // Higher score for longer matches
+    }
+    // Contains match
+    else if (lowerCommandName.includes(lowerQuery)) {
+      matchType = "contains"
+      score = 0.5 + (lowerQuery.length / lowerCommandName.length) * 0.3
+    }
+    // Fuzzy match using Levenshtein distance
+    else {
+      const distance = levenshteinDistance(lowerQuery, lowerCommandName)
```
arseny-kostenko: we can skip Levenshtein when the length difference alone makes it impossible to reach our similarity threshold, e.g. comparing GET (3 chars) with ZRANGEBYSCORE (12 chars) has a minimum edit distance of 9, but our 0.6 similarity cutoff requires distance ≤ 4.8, so we can safely skip that candidate without running the algorithm.

so can add quick wins
```
      const maxLen = Math.max(upperQuery.length, command.name.length)
      const lenDiff = Math.abs(upperQuery.length - command.name.length)
      if (lenDiff > 0.4 * maxLen) continue

      const dist = distance(upperQuery, command.name)
      const similarity = 1 - dist / maxLen
      if (similarity < 0.6) continue
```

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Search commands with fuzzy matching
+ * @param query - Search query
+ * @param options - Search options
+ * @param options.adminMode - Whether to include admin-tier commands
+ * @param options.maxResults - Maximum number of results
+ * @returns Array of match results
+ */
+export function searchCommands(
```
arseny-kostenko: I may be tripping but all these functions are not used in the actual functional code. And some of them are used only in tests. Please remove.

arseny-kostenko: Unnecessary tests too.

apps/frontend/src/utils/valkey-command-matching.ts
```diff
@@ -0,0 +1,200 @@
...
+  return valkeyCommands as ValkeyCommand[]
+}
+
+/**
+ * Get commands filtered by admin mode
+ * @param options - Filter options
+ * @param options.adminMode - Whether to include admin-tier commands (default: false)
+ * @returns Filtered list of commands
+ */
+export function getCommands(options: { adminMode?: boolean } = {}): ValkeyCommand[] {
```
arseny-kostenko: getCommands runs on every search and ultimately runs the filter on every search despite static immutable json. I suggest precomputing these values in the module and export them to be used in `matchCommands`:

```
const [ADMIN_COMMANDS, NON_ADMIN_COMMANDS] = R.partition(R.propEq("tier", "admin"), ALL_COMMANDS)
```

apps/frontend/src/types/valkey-commands.ts
```diff
@@ -0,0 +1,23 @@
...
+  repeatable?: boolean;
+}
+
+export interface ValkeyCommand {
+  name: string;
+  syntax: string;
+  category: string;
+  description: string;
+  parameters: ValkeyParameter[];
+  tier: "default" | "remediation" | "admin";
```
arseny-kostenko: these are very implicit names — it's not straightforward what do they mean. How about "read" | "mutating" | "admin"

apps/frontend/src/utils/valkey-command-matching.ts
allenheltondev: Good call, i just removed this if check because our fallback is an empty array.

apps/frontend/src/utils/valkey-command-matching.ts
allenheltondev: We don't actually have a use case where we need `ADMIN_COMMANDS`. it's either `ALL_COMMANDS` or just `NON_ADMIN_COMMANDS`. So I updated the code appropriately

arseny-kostenko: getCommands is not used

PR #197, opened 2026-01-20
Fix response parsing on metrics server for hotkeys and commandlogs

ravjotbrar:
## Description

### Change Visualization


apps/metrics/docker-compose.yml
```diff
@@ -6,7 +6,9 @@ services:
             CONFIG_PATH: "/app/config.yml"
             PORT: "3000"
             DATA_DIR: "/app/data"
-            VALKEY_URL: "valkey://host.docker.internal:7001" # single URL (router, one seed, or any reachable node)
+            # VALKEY_URL: "valkey://host.docker.internal:7001" # single URL (router, one seed, or any reachable node)
+            VALKEY_HOST: host.docker.internal
+            VALkey_PORT: "7001"
```
arseny-kostenko: it's not all caps `key`

apps/metrics/src/handlers/commandlog-handler.js
```diff
@@ -4,12 +4,14 @@ import * as Streamer from "../effects/ndjson-streamer.js"
 const getCommandLogRows = async (commandlogType, count) => {
   try {
     switch (commandlogType) {
-      case COMMANDLOG_TYPE.SLOW:
```
arseny-kostenko, line 7: all changes here seem unnecessary — it's an async function, so you don't have to await before return

arseny-kostenko (approved):

PR #198, opened 2026-01-20
Add secureStorage for connection passwords

ArgusLi:
## Description

Connection details were stored in plain text in local storage. This change encrypts passwords using Electron's secureStorage before storing, then decrypting it to read.

### DIFF
--- a/apps/frontend/src/state/epics/valkeyEpics.ts
+++ b/apps/frontend/src/state/epics/valkeyEpics.ts
@@ -1,5 +1,5 @@
 import { merge, timer, EMPTY } from "rxjs"
-import { ignoreElements, tap, delay, switchMap, catchError, filter, take, mergeMap } from "rxjs/operators"
+import { ignoreElements, tap, delay, switchMap, catchError, filter, take } from "rxjs/operators"
 import * as R from "ramda"
 import { DISCONNECTED, LOCAL_STORAGE, NOT_CONNECTED, RETRY_CONFIG, retryDelay } from "@common/src/constants.ts"
 import { toast } from "sonner"


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -51,6 +52,25 @@ export const connectionEpic = (store: Store) =>
   merge(
     action$.pipe(
       select(connectPending),
+      mergeMap(async (action) => {
```
arseny-kostenko: two asks: reduce the destructuring noise; reduce nesting by returning early.
```
async (action) => {
  const password = R.path(["payload", "connectionDetails", "password"], action)

  if (R.isNil(password)) return action

  const decryptedPassword = await secureStorage.decrypt(password)

  return assocPath(
    ["payload", "connectionDetails", "password"],
    decryptedPassword,
    action,
  )
}
```

apps/frontend/electron.main.js
```diff
@@ -184,6 +185,22 @@ app.on("activate", () => {
   }
 })
 
+ipcMain.handle("secure-storage:encrypt", async (event, password) => {
+  if (!password || !safeStorage.isEncryptionAvailable()) return password
```
arseny-kostenko, line 189: I don't know... I feel like we should not allow storing plain text password if `!safeStorage.isEncryptionAvailable()`

let's return an empty string if encryption is not available, so the user will have to re-enter the password

apps/frontend/electron.main.js
```diff
@@ -184,6 +185,22 @@ app.on("activate", () => {
...
+  return encrypted.toString("base64")
+})
+
+ipcMain.handle("secure-storage:decrypt", async (event, encryptedBase64) => {
+  if (!encryptedBase64 || !safeStorage.isEncryptionAvailable()) return encryptedBase64
+  try {
+    const encrypted = Buffer.from(encryptedBase64, "base64")
+    return safeStorage.decryptString(encrypted)
+  } catch {
+    return encryptedBase64 // Return as-is if decryption fails
```
arseny-kostenko: can't image anything good to happen after this point — an encrypted, base64 encoded string is no good for anything. you'll display garbage in UI. return an empty string instead and leave a todo here to look at this case more closely

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -65,30 +85,35 @@ export const connectionEpic = (store: Store) =>
...
+            }
+          }
+
+          const connectionToSave = {
+            connectionDetails,
+            status: NOT_CONNECTED,
+            connectionHistory: connection?.connectionHistory || [],
+          }
+          currentConnections[payload.connectionId] = connectionToSave
+          localStorage.setItem(LOCAL_STORAGE.VALKEY_CONNECTIONS, JSON.stringify(currentConnections))
```
arseny-kostenko, line 105: see — so if encryption failed — you will store a plain text password

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -65,30 +85,35 @@ export const connectionEpic = (store: Store) =>
           type === standaloneConnectFulfilled.type ||
           type === clusterConnectFulfilled.type,
       ),
-      tap(({ payload }) => {
+      mergeMap(async ({ payload }) => {
```
arseny-kostenko: it's a side-effect, why mergeMap returning EMPTY instead of tap?

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -65,30 +85,35 @@ export const connectionEpic = (store: Store) =>
...
-            (p) => ({
-              connectionDetails: connection?.connectionDetails || p.connectionDetails,
-              status: NOT_CONNECTED,
-              connectionHistory: connection?.connectionHistory || [],
-            }),
-            (connectionToSave) => ({ ...currentConnections, [payload.connectionId]: connectionToSave }),
-            JSON.stringify,
-            (updated) => localStorage.setItem(LOCAL_STORAGE.VALKEY_CONNECTIONS, updated),
-          )(payload)
+          let connectionDetails = connection?.connectionDetails || payload.connectionDetails
```
arseny-kostenko: There is no good justification for `let` and a reassignment. You can simply:
```
const baseConnectionDetails =
  connection?.connectionDetails ?? payload.connectionDetails

const connectionToSave = {
  connectionDetails: R.isNil(R.path(["password"], baseConnectionDetails))
    ? baseConnectionDetails
    : R.assoc(
        "password",
        await secureStorage.encrypt(baseConnectionDetails.password),
        baseConnectionDetails,
      ),
  status: NOT_CONNECTED,
  connectionHistory: connection?.connectionHistory ?? [],
}

currentConnections[payload.connectionId] = connectionToSave
```

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -250,25 +275,35 @@ export const deleteConnectionEpic = () =>
...
+    mergeMap(async ({ payload: { connectionId } }) => {
       try {
         const currentConnections = getCurrentConnections()
 
         const state = store.getState()
         const connection = state.valkeyConnection?.connections?.[connectionId]
 
         if (connection && currentConnections[connectionId]) {
-          currentConnections[connectionId].connectionDetails = connection.connectionDetails
+          let connectionDetails = connection.connectionDetails
```
arseny-kostenko: same as above:
```
const connectionDetails = connection.connectionDetails.password
  ? R.assoc(
      "password",
      await secureStorage.encrypt(connection.connectionDetails.password),
      connection.connectionDetails,
    )
  : connection.connectionDetails
  ```

apps/frontend/src/state/epics/valkeyEpics.ts
ArgusLi: Leftover from a change that resulted in an error.

Issue #206, opened 2026-01-28
Can still access routes that require connection when valkey server is disconnected


ravjotbrar:
When the WebSocket disconnects due to the client machine going to sleep, routes that require an active connection remain clickable. This is confusing to users because there is no clear indication that the client-server connection has been lost.

<img width="1197" height="794" alt="Image" src="https://github.com/user-attachments/assets/7c26335d-abee-4749-b99d-e15a77f72cc1" />

Steps to reproduce:

1. Connect to valkey instance and go to connections page
2. Put computer to sleep for a small amount of time (e.g. 10s)
3. Wake up computer.

This can also be reproduced in other routes if we sleep the entire duration of the reconnect attempts.

Expected behaviour:

- Server reconnects automatically, prompting the client/s to also attempt a reconnect

PR #211, opened 2026-01-31
Fix WS and Valkey reconnects 

ravjotbrar:
## Description

This change fixes a few different issues with our reconnects. The first is that reconnects were still being attempted even after the system went to sleep. This resulted in a unrecoverable connection error for the ws and valkey connections. This was solved by sending a message from electron to our backend when the system was going to sleep and when it woke up and pausing/resuming retries accordingly.
https://github.com/valkey-io/valkey-admin/issues/213

The second issue was that multiple websockets were being recreated upon reconnection to the websocket server, resulting in a valkey reconnection being stuck in a `connecting` state and giving the user no way to connect except for deleting the connection through the UI and retrying manually. The fix for this was to use exhaustMap in our wsEpic instead of mergeMap to force a single websocket at a time.
https://github.com/valkey-io/valkey-admin/issues/212

The third biggest issue this change addresses is still being able access routes that require connection when there is no active connection. This is solved by removing the RequireConnection condition for the ValkeyReconnect route.
https://github.com/valkey-io/valkey-admin/issues/206

Other minor fixes:
- Rename file `rxjsMiddlware` to `rxjsMiddleware`
- Remove isDisconnected conditional for the connect button as it is redundant
- Ensure our useIsConnected hook returns only when the status of a connection is `CONNECTED`
- Filtered on connection errors before sending a connectRejected due to failed valkey actions

### DIFF
--- a/apps/frontend/src/components/connection/ConnectionEntry.tsx
+++ b/apps/frontend/src/components/connection/ConnectionEntry.tsx
@@ -132,7 +132,7 @@ export const ConnectionEntry = ({
                 </Tooltip>
               </>
             )}
-            {((!isConnected && !isConnecting)) && (
+            {(!isConnected && !isConnecting) && (
               <Tooltip>
                 <TooltipTrigger asChild>
                   <Button onClick={handleConnect} size="sm" variant="ghost">
@@ -232,7 +232,7 @@ export const ConnectionEntry = ({
               </Tooltip>
             </>
           )}
-          {((!isConnected && !isConnecting)) && (
+          {(!isConnected && !isConnecting) && (
             <Tooltip>
               <TooltipTrigger asChild>
                 <Button onClick={handleConnect} size="sm" variant="ghost">
--- a/apps/frontend/src/state/epics/wsEpics.ts
+++ b/apps/frontend/src/state/epics/wsEpics.ts
@@ -20,7 +20,8 @@ import {
   connectPending,
   connectRejected,
   reconnectAttempt,
-  reconnectExhausted
+  reconnectExhausted,
+  selectRetriesPaused
 } from "@/state/wsconnection/wsConnectionSlice"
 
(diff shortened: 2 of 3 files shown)


apps/frontend/src/components/connection/ConnectionEntry.tsx
```diff
@@ -142,7 +132,7 @@ export const ConnectionEntry = ({
                 </Tooltip>
               </>
             )}
-            {(isDisconnected || (!isConnected && !isConnecting)) && (
+            {((!isConnected && !isConnecting)) && (
```
arseny-kostenko: too many braces?

apps/frontend/src/components/connection/ConnectionEntry.tsx
```diff
@@ -242,7 +232,7 @@ export const ConnectionEntry = ({
               </Tooltip>
             </>
           )}
-          {(isDisconnected || (!isConnected && !isConnecting)) && (
+          {((!isConnected && !isConnecting)) && (
```
arseny-kostenko: too many braces?

apps/frontend/src/state/epics/wsEpics.ts
```diff
@@ -71,21 +74,24 @@ const connect = (store: Store) =>
...
             store.dispatch(reconnectAttempt({
               attempt: retryCount,
               maxRetries: RETRY_CONFIG.MAX_RETRIES,
               nextRetryDelay: delay,
             }))
-
-            return timer(delay)
+            return action$.pipe(
+              filter((action) =>
+                !store.getState()["websocket"].reconnect.retriesPaused || action.type === "websocket/resumeRetries",
```
arseny-kostenko: should create a selector for this https://redux-toolkit.js.org/api/createSelector

PR #217, opened 2026-02-09
improve the cluster topology view to accomodate more nodes and improv…

nassery318:
…ed the send command

## Description

1. Improved the cluster topology from card view to row view with primary (details) and replica (details)
2. Refactored the send command view with ui primitives

### Change Visualization

https://github.com/user-attachments/assets/8dac98b0-ca06-4180-b5a2-72552953675c

### DIFF
--- a/apps/frontend/src/components/cluster-topology/Cluster.tsx
+++ b/apps/frontend/src/components/cluster-topology/Cluster.tsx
@@ -49,29 +49,16 @@ export function Cluster() {
   // filtering nodes based on search query
   const filteredEntries = clusterEntries.filter(([primaryKey, primary]) => {
     if (!searchQuery) return true
-    const query = searchQuery.toLowerCase()
-    const primaryData = clusterData.data[primaryKey]
 
     // check primary node
-    if (
-      primaryKey.toLowerCase().includes(query) ||
-      primary.host.toLowerCase().includes(query) ||
-      primary.port.toString().includes(query) ||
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -24,22 +36,102 @@ export function Cluster() {
...
+  const connectedNodes = clusterEntries.filter(([primaryKey]) =>
+    connectionStatuses[primaryKey]?.status === CONNECTED,
+  ).length
+  const totalReplicas = clusterEntries.reduce((sum, [, primary]) =>
+    sum + primary.replicas.length, 0,
+  )
+  const totalClusterNodes = totalNodes + totalReplicas
+
+  // filtering nodes based on search query
+  const filteredEntries = clusterEntries.filter(([primaryKey, primary]) => {
```
arseny-kostenko, line 50: there is a lot of work going on which may become noticeable on larger clusters with hundreds of nodes, so a couple of suggestions:
1. Precompute a searchableText per node in the clusterSlice (maybe in setClusterData) so that you don't have to run multiple `.includes` and `toLowerCase`
2. Instead of `searchQuery.toLowerCase()` in each iteration of `filter`, do it at the call site `onChange={(e) => setSearchQuery(e.target.value.toLowerCase())}`

arseny-kostenko (approved):

PR #219, opened 2026-02-11
Limit connections to 12 and fix application connection and deletion logic

ravjotbrar:
## Description

- Start one metrics server for only connected nodes, instead of all cluster nodes
- Close metrics server when we disconnect/delete a connection by adding a delete endpoint 
- Limit active connections to a maximum of 12
- Change tap in connection epic to mergeMap so returns aren't ignored
- Fix delete epic to close and delete connections appropriately
- Reuse GlideClusterClient for all nodes belonging to the same cluster
- Add clusterId to clients map so we know when to close glide cluster client
- Add clusterNodesMap to keep track of connected nodes per cluster

### Change Visualization
Limited to 1 for testing purposes

https://github.com/user-attachments/assets/13e7d0a9-f4f8-4938-a605-978d7cbdcc63

### DIFF
--- a/apps/metrics/src/index.js
+++ b/apps/metrics/src/index.js
@@ -119,6 +119,8 @@ async function main() {
       const { connectionId } = req.body
       client.close()
       const ownConnectionId = sanitizeUrl(`${process.env.VALKEY_HOST}-${Number(process.env.VALKEY_PORT)}`)
+      // Required to kill electron-managed metrics service
+      // TODO: revisit to decouple concerns
       if (connectionId === ownConnectionId) process.send?.({ type: "close-client", payload: { connectionId } })
       return res.status(200).json({
         ok: true,


apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -113,11 +115,18 @@ export function ClusterNode({
 
             {/* Actions */}
             <div className="flex items-center gap-2 shrink-0">
-              <CustomTooltip content={`${isConnected ? "Connected" : "Not Connected"}`}>
+              <CustomTooltip content={`${isConnected ? "Connected" : isDisabled ? `Max connections of ${MAX_CONNECTIONS} reached` : "Not Connected" }`}>
                 <PowerIcon
-                  className={`${isConnected ? "text-green-500 bg-green-100" : 
-                    "text-gray-400 cursor-pointer bg-gray-100 hover:text-gray-600"} rounded-full p-0.5`}
-                  onClick={handleNodeConnect}
+                  className={`
```
arseny-kostenko: I recommend using `cn`
```
className={cn(
  "rounded-full p-0.5",
  isConnected && "text-green-500 bg-green-100",
  !isConnected && isDisabled && "text-gray-300 cursor-not-allowed bg-gray-100",
  !isConnected && !isDisabled && "text-gray-400 cursor-pointer bg-gray-100 hover:text-gray-600"
)}
```

apps/frontend/src/components/ui/connection-form.tsx
```diff
@@ -23,7 +25,8 @@ function ConnectionForm({ onClose }: ConnectionFormProps) {
     alias: "",
   })
   const [connectionId, setConnectionId] = useState<string | null>(null)
-
+  const connectedConnections = useSelector(selectConnectionCount)
+  const isAtConnectionLimit = connectedConnections >= MAX_CONNECTIONS
```
arseny-kostenko: since it's the second occurrence, you can create a new derived selector and use it instead
```
export const selectIsAtConnectionLimit = (state) => selectConnectionCount(state) >= MAX_CONNECTIONS
```

apps/frontend/src/state/valkey-features/connection/connectionSelectors.ts
```diff
@@ -1,11 +1,13 @@
...
-import { VALKEY } from "@common/src/constants.ts"
+import { VALKEY, CONNECTED } from "@common/src/constants.ts"
 import type { RootState } from "@/store.ts"
 
 export const atId = R.curry((id: string, state: RootState) => R.path([VALKEY.CONNECTION.name, "connections", id], state))
 
 export const selectStatus = (id: string) => (state: RootState) => atId(id, state)?.status
 export const selectConnectionDetails = (id: string) => (state: RootState) => atId(id, state)?.connectionDetails
 export const selectConnections = (state: RootState) => state[VALKEY.CONNECTION.name].connections
+export const selectConnectionCount = (state: RootState) => Object.values(state[VALKEY.CONNECTION.name].connections).filter(
```
arseny-kostenko: you should reuse `selectConnections` from above

apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
```diff
@@ -211,15 +211,22 @@ const connectionSlice = createSlice({
       state.connections[connectionId].status = NOT_CONNECTED
       state.connections[connectionId].errorMessage = null
     },
+    closeConnectionFulfilled: () => {
+
```
arseny-kostenko, line 219: do we not do anything here? maybe status = DISCONNECTED?

apps/metrics/src/index.js
```diff
@@ -113,6 +113,24 @@ async function main() {
     }
   })
 
+  app.delete("/close-client/:connectionId", async (req, res) => {
```
arseny-kostenko: `:connectionId` is unnecessary — it's one :connectionId per metrics service. We only need to ensure that the electron app hits the right one

arseny-kostenko: `delete` is misleading. We're not deleting a real resource like a row from a db, we're performing an operation. I would suggest POST `/connection/close` (so we can also add `/connection/connect` and `/connection/status` if needed)

arseny-kostenko: `connectionId` can still be put in the payload for validation, so further in the code, before you `send`, you do a validation
```
const ownConnectionId = sanitizeUrl(`${process.env.VALKEY_HOST}-${Number(process.env.VALKEY_PORT)}`)
if (connectionId === ownConnectionId) process.send?.({ type: "close-client", payload: { connectionId } })
```

apps/server/src/actions/config.ts
```diff
@@ -50,8 +50,8 @@ export const enableClusterSlotStats = withDeps<Deps, void>(
   async ({ clients, action }) => {
     const { connectionIds } = action.payload
     const promises = connectionIds.map(async (connectionId: string) => {
-      const client = clients.get(connectionId)
-      await client?.customCommand(["CONFIG", "SET", "cluster-slot-stats-enabled", "yes"])
+      const connection = clients.get(connectionId)
+      await connection?.client?.customCommand(["CONFIG", "SET", "cluster-slot-stats-enabled", "yes"])
```
arseny-kostenko, line 58: now that we don't spin up metrics services for every node, this will set cluster-slot-stats-enabled only for connected nodes, which is probably not a good thing. We might want to run here `CLUSTER SHARDS` first, then iterate over all nodes (maybe even using a single client instance) and set it on all nodes

apps/server/src/actions/connection.ts
```diff
@@ -36,3 +37,17 @@ export const resetConnection = withDeps<Deps, void>(
...
   },
 )
+
+export const closeConnection = withDeps<Deps, void>(
+  async ({ ws, clients, action, metricsServerURIs }) => {
+    const { connectionId } = action.payload 
+    const connection = clients.get(connectionId)
+    closeMetricsServer(connectionId, metricsServerURIs)
+    if (connection && (connection.client instanceof GlideClient
+      || (connection.client instanceof GlideClusterClient && await isLastConnectedClusterNode(connectionId, clients)))
```
arseny-kostenko: let's move this `await isLastConnectedClusterNode` inside the `if` and make it another `if` and leave a comment about why this check matters

apps/server/src/connection.ts
```diff
@@ -107,16 +107,10 @@ async function discoverCluster(client: GlideClient, payload: {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const response = await client.customCommand(["CLUSTER", "SLOTS"]) as any[][]
 
-    const { clusterId } = R.applySpec({
-      firstSlotRange: R.nth(0),
-      firstPrimaryNode: R.path([0, 2]), // firstSlotRange[2]
-      clusterId: R.path([0, 2, 2]), // firstPrimaryNode[2]
-    })(response)
+    const clusterId = R.path([0, 2, 2], response)
```
arseny-kostenko, line 120: leave a comment what this magical 0, 2, 2 sequence is

apps/server/src/connection.ts
```diff
@@ -164,49 +158,66 @@ async function discoverCluster(client: GlideClient, payload: {
...
+  const existingKey = Object.keys(clusterNodes).find(
+    (key) => clients.get(key) instanceof GlideClusterClient,
+  )
+
+  const existingConnection = existingKey
+    ? clients.get(existingKey)
+    : undefined
+
+  if (existingConnection) {
+    const existingClient = existingConnection.client
```
arseny-kostenko: nit: destructuring would be easier to read
```
const { client: existingClient, clusterId: existingClusterId } = existingConnection
```

apps/server/src/connection.ts
```diff
@@ -231,16 +242,53 @@ async function connectToCluster(
...
-  clients: Map<string, GlideClient | GlideClusterClient>,
+  clients: Map<string, {client: GlideClient | GlideClusterClient, clusterId?: string}>,
   ws: WebSocket) 
 {
-  const resolvedAddresses = (await resolveHostnameOrIpAddress(payload.connectionDetails.host)).addresses
-  if (resolvedAddresses.some((address) => clients.has(sanitizeUrl(`${address}:${payload.connectionDetails.port}`)))) {
+  const { connectionId, connectionDetails } = payload
+  const resolvedAddresses = (await resolveHostnameOrIpAddress(connectionDetails.host)).addresses
+  if (resolvedAddresses.some((address) => clients.has(sanitizeUrl(`${address}:${connectionDetails.port}`)))
+      || (clients.has(connectionId) && clients.get(connectionId) instanceof GlideClient))  {
```
arseny-kostenko, line 266: we're doing this instanceof GlideClient check in different places — is it even possible to be something else? is this unnecessary defensive?

apps/server/src/connection.ts
```diff
@@ -231,16 +242,53 @@ async function connectToCluster(
...
+        payload: { connectionId },
       }),
     )
   }
 }
+
+export async function closeMetricsServer(connectionId: string, metricsServerURIs: Map<string, string>) {
+  const metricsServer = metricsServerURIs.get(connectionId)
+  if (metricsServer) {
+    const res = await fetch(`${metricsServer}/close-client/${encodeURIComponent(connectionId)}`, { method: "DELETE" })
```
arseny-kostenko: see above in metrics

apps/server/src/connection.ts
```diff
@@ -231,16 +242,53 @@ async function connectToCluster(
...
+    if (res.ok) console.log("Connection closed successfully")
```
arseny-kostenko: `Connection closed successfully ${connectionId}`

apps/server/src/connection.ts
```diff
@@ -231,16 +242,53 @@ async function connectToCluster(
...
+    else console.warn("Could not kill metrics server process")
+  }
+}
+
+export async function closeClient(
+  connectionId: string,
+  client: GlideClient | GlideClusterClient | undefined,
+  ws: WebSocket,
+)  {
+  if ( client) {
```
arseny-kostenko: extra space

common/src/constants.ts
```diff
@@ -96,6 +98,9 @@ export const ERROR = "Error"
 export const NOT_CONNECTED = "Not Connected"
 export const DISCONNECTED = "Disconnected"
 export const RECONNECTING = "Reconnecting"
+export const DELETING = "Deleting"
+// Change to 12 after testing
```
arseny-kostenko: Image

apps/server/src/actions/config.ts
ravjotbrar: This uses a GlideClusterClient which will enable it for the entire cluster

apps/server/src/actions/connection.ts
ravjotbrar: even though both conditions result in closeClient?

apps/server/src/connection.ts
ravjotbrar: GlideClusterClient is the other option. It differentiates between standalone instances and cluster instances. In this case, I may need to add functionality to return if duplicate cluster node connections as well

apps/server/src/actions/connection.ts
arseny-kostenko: this `if` is hard to read as it's doing too much, and this `await` inside `if` is not tasteful. Alternatively, you can add a helper 
```
const canSafelyDisconnect = async (
  connectionId: string,
  connection: { client: GlideClient | GlideClusterClient } | undefined,
  clients: Map<string, any>
) => {
  if (connection?.client instanceof GlideClient) return true

  if (connection?.client instanceof GlideClusterClient)
    return isLastConnectedClusterNode(connectionId, clients)

  return false
}
```

apps/server/src/actions/config.ts
arseny-kostenko: then we're potentially repeating ourselves as it should do it once per clusterId but here it can potentially do it against several nodes of the same cluster?

apps/server/src/connection.ts
arseny-kostenko: yeah, I see. It's another heavy `if`. Let's, at least, leave a human readable concise comment

apps/server/src/actions/config.ts
ravjotbrar: Technically, this functionality isn't exposed yet and we may just discard it entirely. The idea was to have some sort of button that would allow users to enable cluster slot stats from the UI. Not sure we need that given that they could enable it from SendCommand anyways.

apps/server/src/actions/connection.ts
ravjotbrar: Fair enough, will change

apps/frontend/src/state/valkey-features/connection/connectionSlice.ts
ravjotbrar: On second thought, yes. I changed the status to NOT_CONNECTED and will at a DISCONNECTING state for closeConnection action.

apps/metrics/src/index.js
ravjotbrar: Changed to POST, connection/close, and derive connectionId from env variables. The only thing I didn't add is checking ownConnectionId against connectionId, because I don't know where I would get connectionId from if not passed through api call.

arseny-kostenko: connectionId  — in POST body, or header. My point was not about not passing a connectionId from server to metrics, but about not making connectionId part of URL because connection is not a resource

apps/server/src/actions/config.ts
arseny-kostenko: what I suggest is change the payload from 
```
const { connectionIds } = action.payload
```
to
```
const { clusterIds } = action.payload
```
and then map clusterIds to send command.

I do realize that we map glide clients by connectionId, so we need to add a new map <connectionId, clusterId> for a quick lookup of a clusterId by connectionId. But given that we have this concept that cluster-level commands can be routed through a single node, we may want to have an easy way track commands against a cluster, not just a node

arseny-kostenko: or we can keep the payload the same but calculate the clusterIds only on the backend

apps/metrics/src/index.js
```diff
@@ -113,6 +114,25 @@ async function main() {
     }
   })
 
+  app.post("connection/close", async (req, res) => {
+    try {
+      const { connectionId } = req.body
+      client.close()
+      const ownConnectionId = sanitizeUrl(`${process.env.VALKEY_HOST}-${Number(process.env.VALKEY_PORT)}`)
+      if (connectionId === ownConnectionId) process.send?.({ type: "close-client", payload: { connectionId } })
```
arseny-kostenko, line 124: this code has implicit meaning and can evoke wtfs from a reader not intimately familiar with the codebase. Let's leave a comment that this is specifically for electron-managed metrics service, so that electron can kill the process. Now when I look at it, it almost feels like a hack/tight coupling of several concerns (disconnecting and killing the process), so maybe let's leave a todo as well to revisit this

arseny-kostenko (approved):

PR #228, opened 2026-02-23
added typography and applied it to connection, dashboard and primitives

nassery318:
## Description

1. Added custom Inter font
2. Define named typography roles — text-heading, text-body-xs, text-code
3. Applied the typography using a centralized Typography component 

### Change Visualization

image

### DIFF
--- a/apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
+++ b/apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
@@ -158,7 +158,6 @@ export const ClusterConnectionGroup = ({ clusterId, connections, onEdit }: Clust
                   ) : (
                     <div className="min-w-0 flex-1 overflow-hidden max-w-[200px]">
                       <Typography
-                        // as="h3"
                         className="hover:underline block truncate"
                         onClick={() => setIsOpen(!isOpen)}
                         title={firstNodeAlias || clusterId}


apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
```diff
@@ -137,19 +138,23 @@ export const ClusterConnectionGroup = ({ clusterId, connections, onEdit }: Clust
...
-                            to={`/${clusterId}/${firstConnectedConnection.connectionId}/cluster-topology`}
+                          <Typography
+                            variant="code"
                           >
-                            {firstNodeAlias || clusterId}
-                          </Link>
+                            <Link
+                              className="hover:underline"
+                              style={{
+                                overflow: "hidden",
```
arseny-kostenko: why not className `block truncate`?

apps/frontend/src/components/ui/typography.tsx
```diff
@@ -0,0 +1,67 @@
...
+  heading: "h3",
+  subheading: "h4",
+  label: "p",
+  body: "p",
+  bodyLg: "p",
+  bodySm: "p",
+  bodyXs: "p",
+  caption: "span",
+  overline: "span",
+  code: "code",
```
arseny-kostenko, line 46: `code` is inline, so I guess we're missing a code block that should come with `pre`

apps/frontend/src/css/index.css
```diff
@@ -47,6 +48,10 @@
   --color-tw-dark-border: var(--tw-dark-border);
   --color-tw-accent: var(--tw-accent);
   --color-tw-accent2: var(--tw-accent2);
+
+  /* Font */
+  --font-sans: 'Inter Variable', system-ui, -apple-system, 'Segoe UI', sans-serif;
+  --font-mono: ui-monospace, 'SF Mono', Consolas, monospace;
```
arseny-kostenko: maybe we need a custom monospace font too? ui-monospace will be an OS's font which will render differently on Windows and Mac and Linux, i.e. some of them are larger than others

apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
nassery318: removed inline styles.

apps/frontend/src/components/ui/typography.tsx
nassery318: codeBlock added for pre

apps/frontend/src/css/index.css
nassery318: added the custom geist-mono for keeping the mono text consistent for different OSs.

PR #231, opened 2026-02-26
Limit key size fetched by key browser to 2KB

ArgusLi:
## Description

For keys with size >2KB, do not fetch their value in key browser, instead, return a message stating that keys above 2KB are not displayed in the UI.

### Change Visualization

image

### DIFF
--- a/apps/server/src/keys-browser.ts
+++ b/apps/server/src/keys-browser.ts
@@ -11,6 +11,7 @@ import {
 import pLimit from "p-limit"
 import { VALKEY, VALKEY_CLIENT } from "../../../common/src/constants.ts"
 import { buildScanCommandArgs } from "./valkey-client-commands.ts"
+import { formatBytes } from "../../../common/src/bytes-conversion.ts"
 
 interface EnrichedKeyInfo {
   name: string;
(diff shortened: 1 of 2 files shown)


common/src/constants.ts
```diff
@@ -130,6 +130,7 @@ export const VALKEY_CLIENT = {
     defaultPayloadPattern: "*",
     defaultCount: 50,
   } ,
+  KEY_VALUE_SIZE_LIMIT: 2000, // 2KB
```
arseny-kostenko: I believe, this is not correct. I kilobytes is not 1000 bytes, it's 2^10 = 1024

apps/server/src/keys-browser.ts
```diff
@@ -60,6 +60,15 @@ export async function getKeyInfo(
 
       const commands = elementCommands[keyType.toLowerCase()]
       if (commands) {
+        if (memoryUsage > VALKEY_CLIENT.KEY_VALUE_SIZE_LIMIT) {
+          if (commands.sizeCmd){
+            keyInfo.collectionSize = await (client.customCommand([commands.sizeCmd, key])) as number
+          }
+          keyInfo.elements = `Keys above ${VALKEY_CLIENT.KEY_VALUE_SIZE_LIMIT / 1000}KB are not displayed in UI.`
```
arseny-kostenko: let's provide the size of the key since we already have it

common/src/constants.ts
ArgusLi: As a SI unit, 1 KB (kilobyte) is 1000 bytes. Kibibyte is 1024 bytes. I'm happy to use Kibibyte, just not sure which to use.

ArgusLi: Looking at the UI for key browser, we are using formatBytes, which uses Kibibytes. I'll make the change.

PR #234, opened 2026-03-02
truncate long commands and show details with tooltip

nassery318:
## Description

Long commands in the command logs table was distorting the alignment of table therefore below fixes were made:

- Truncate long commands
- Show the full command on a tooltip

### Change Visualization

https://github.com/user-attachments/assets/1428be37-abe1-400e-846f-f2d19931fa07

### DIFF
--- a/apps/frontend/src/components/monitoring/command-log-table.tsx
+++ b/apps/frontend/src/components/monitoring/command-log-table.tsx
@@ -6,6 +6,7 @@ import { EmptyState } from "../ui/empty-state"
 import { TableContainer } from "../ui/table-container"
 import { SortableTableHeader, StaticTableHeader } from "../ui/sortable-table-header"
 import { Typography } from "../ui/typography"
+import { CustomTooltip } from "../ui/tooltip"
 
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/monitoring/command-log-table.tsx
```diff
@@ -132,9 +139,14 @@ export function CommandLogTable({ data, logType }: CommandLogTableProps) {
...
             <td className="px-4 py-2 flex-1">
-              <Typography className="bg-primary/30 py-1 px-2 rounded-full"  variant="code">
-                {entry.argv.join(" ")}
-              </Typography>
+              <CustomTooltip content={entry.argv.join(" ")}>
+                <Typography
+                  className="bg-primary/30 py-1 px-2 rounded-full"
+                  variant="code"
+                >
+                  {truncateCommand(entry.argv)}
```
arseny-kostenko, line 147: curious, if tailwind `truncate` doesn't solve this

nassery318: image

I tried truncate, however it does not give access to the character length. Above is the result with truncate only.

PR #238, opened 2026-03-10
Replace non-human-readable keys and values with placeholder text

ArgusLi:
## Description

When Valkey stores binary or non-UTF-8 data, the Key Browser can display garbled, non-human-readable strings. This PR detects such content and replaces it with a "Not human readable" message instead.

### Human-readability detection

A string is considered human-readable if at least 90% of its characters are printable ASCII (code points 32–126, plus tab, line feed, and carriage return). A ratio-based approach is used instead of a strict all-or-nothing check because some otherwise readable strings can contain stray non-printable bytes (e.g. a trailing null byte or minor encoding artifact). The 90% threshold cleanly separates "mostly text" content from truly binary data, which typically has a printable ratio well below 50%.

### Change Visualization

image

### DIFF
--- a/apps/frontend/src/components/key-browser/KeyBrowser.tsx
+++ b/apps/frontend/src/components/key-browser/KeyBrowser.tsx
@@ -155,7 +155,7 @@ export function KeyBrowser() {
             className="flex-1"
             label="Total Keys"
             tooltip={
-              <TooltipIcon description="Total number of keys in the database" size={14} />
+              <TooltipIcon description="Total number of keys sampled" size={14} />
             }
             value={totalKeys}
           />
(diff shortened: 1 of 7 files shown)


apps/server/src/utils.ts
```diff
@@ -141,3 +142,48 @@ export async function isLastConnectedClusterNode(
...
+export const isHumanReadable = (str: string): boolean => {
+  if (str.length === 0) return true;
+
+  let printableCount = 0;
+
+  for (const char of str) {
+    const code = char.charCodeAt(0)
+
+    if (
+          (code >= 32 && code <= 126) // letters, numbers, punctuations
```
ravjotbrar: Hmm what about unicode?

ArgusLi: This implementation only considers ASCII printable characters and doesn't consider any non-english characters.

I'll make the change to fix this.

arseny-kostenko (dismissed):

arseny-kostenko (approved):

PR #239, opened 2026-03-10
added alert primitive and used it in components

nassery318:
## Description

1. Added alert primitive component and reused across the app for consistency 
2. Other ui improvements like tooltips, search and etc..

### Change Visualization

https://github.com/user-attachments/assets/baf7527f-ca28-46a7-beba-d74b40c9aa92

### DIFF
--- a/apps/frontend/src/components/key-browser/KeyBrowser.tsx
+++ b/apps/frontend/src/components/key-browser/KeyBrowser.tsx
@@ -155,7 +155,7 @@ export function KeyBrowser() {
             className="flex-1"
             label="Total Keys"
             tooltip={
-              <TooltipIcon description="Total number of keys scanned/fetched" size={14} />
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/key-browser/KeyBrowser.tsx
```diff
@@ -155,15 +155,15 @@ export function KeyBrowser() {
             className="flex-1"
             label="Total Keys"
             tooltip={
-              <TooltipIcon description="Total number of keys in the database" size={14} />
+              <TooltipIcon description="Total number of keys scanned/fetched" size={14} />
```
ArgusLi: Would "sampled" be a better word here than "scanned/fetched"?

nassery318: Definitely, sampled is much better. Thanks

PR #240, opened 2026-03-12
Add metrics files cleanup

ArgusLi:
## Description

Cleans metrics files older than a set retention days period. Cleanup service runs on metrics launch, and every 24 hours after.

### DIFF
--- a/apps/metrics/src/effects/ndjson-cleaner.js
+++ b/apps/metrics/src/effects/ndjson-cleaner.js
@@ -12,14 +12,17 @@ let cleanerStopper
 export const setupNdjsonCleaner = ( cfg ) => {
   const pipeline$ = timer(0, METRICS_EVICTION_POLICY.INTERVAL).pipe(
     exhaustMap(() => 
-      fs.promises.readdir(cfg.server.data_dir)
+      fs.promises.readdir(cfg.server.data_dir) // get file names from directory
+        // get file stats for each file
         .then((fileNames) => Promise.all(fileNames.map(
           async (fileName) => ({ stats: await fs.promises.stat(path.join(cfg.server.data_dir, fileName)), fileName }),
         )))
+        // filter out files that are not expired
         .then((filesWithStats) => filesWithStats.reduce((acc, { stats, fileName }) => { 
           if (stats.birthtime < Date.now() - cfg.storage.retention_days * MILLISECONDS_IN_A_DAY) acc.push(fileName)
           return acc
         }, []))
+        // delete expired files
         .then(async (expiredFileNames) => (await Promise.allSettled(expiredFileNames.map(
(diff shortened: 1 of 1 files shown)


apps/metrics/src/effects/ndjson-cleaner.js
```diff
@@ -0,0 +1,59 @@
...
+
+const applyRetentionPolicy = (fileNames, retentionDays) => {
+  // Calculate cutoff date in UTC, aligning cutoff to midnight.
+  const now = Date.now()
+  const todayUTC = now - (now % MILLISECONDS_IN_A_DAY)
+  const cutoff = todayUTC - retentionDays * MILLISECONDS_IN_A_DAY
+
+  const expired = fileNames.filter((fileName) => {
+    if (!fileName.endsWith(".ndjson")) return false
+    const match = fileName.match(/_(\d{8})\.ndjson$/)
```
arseny-kostenko: I'm pretty sure this is going to bite you in the soft tissues in the future — you're making a huge assumption that this regex will work and that the date will stay the encoded in the filename in this specific way. I guarantee that this will break when you (or somebody else) start working on rotating files by size so you will have more than 8 digits or additional extensions. Use fs.stat and maybe atime as the only real immutable timestamp

apps/metrics/src/effects/ndjson-cleaner.js
```diff
@@ -0,0 +1,59 @@
...
+    if (!match) return false
+    const [y, m, d] = [match[1].slice(0, 4), match[1].slice(4, 6), match[1].slice(6, 8)]
+    // Month uses month index instead of month number
+    return Date.UTC(y, m - 1, d) < cutoff
+  })
+  return expired
+}
+
+const deleteFiles = async (dir, fileNames) => {
+  await Promise.all(fileNames.map((fileName) => fs.promises.unlink(path.join(dir, fileName))))
```
arseny-kostenko: I think, here we need allSettled. If one file wasn't deleted for some reason, we shouldn't fail the entire batch

apps/metrics/src/effects/ndjson-cleaner.js
ArgusLi: That makes so much more sense. I didn't even think about it...

apps/metrics/config.yml
```diff
@@ -10,6 +10,9 @@ collector:
   batch_ms: 60000
   batch_max: 500
 
+storage:
+  retention_days: 30
```
arseny-kostenko: let's do 5 — 30 can run out of disk space

apps/metrics/src/effects/ndjson-cleaner.js
arseny-kostenko: and then iterate over results and log errors if any so we don't fail deletion silently

PR #241, opened 2026-03-13
Web Based Support

ravjotbrar:
## Description

This PR adds web support to our entire application by adding a metrics orchestrator service in the backend to start and stop metrics servers. The metrics orchestrator has two endpoints:

- Register -> gives the metrics server a way to register itself and pass its URI to the backend
- Ping -> gives the metrics server a way to convey its health

Currently, the metrics orchestrator has two modes of operation. The first is similar to how we created metrics servers in the desktop application using electron; i.e. forking a child process every time we connected to a Valkey node. The second mode requires an environment variable (to-be-named) to fork metrics server processes for _each_ node belonging to an initial cluster node that is provided by the user (also through env variables). The flow of this mode works as follows:

1. Connect to Valkey using initial node details 
2. Discover cluster topology if that initial node belongs to a cluster
3. Start metrics servers for every node

We periodically receive pings from the metrics servers, which update a lastSeen property in our metrics map. If we haven't seen a node (received a ping) for a specified amount of time, we try to kill that process.
 
We also periodically update cluster map by discovering cluster topology to detect changes in nodes. We then start metrics servers for any newly discovered nodes and stop metrics servers for any nodes that no longer belong to the cluster.

Aside from this change, many of the smaller changes in the imports and packages are concerning the building of our application so that it is compatible for desktop and web.


apps/metrics/src/index.js
```diff
@@ -142,10 +143,56 @@ async function main() {
...
-    process.send?.({ type: "metrics-started", payload: { metricsHost: "http://0.0.0.0", metricsPort: assignedPort } })
+    try {
+      const registerURI = `http://${backendServerHost}:${backendServerPort}/orchestrator/register`
+      console.debug("Sending Register request to ", registerURI)
+      const response = await fetch(registerURI,
+        {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            metricsServerUri: `http://0.0.0.0:${assignedPort}`,
```
arseny-kostenko: let's replace 0.0.0.0 with an env variable that defaults to 127.0.0.1 — this is aligned with the security findings

common/src/json-utils.ts
```diff
@@ -58,15 +58,15 @@ export const diff = (a:JSONObject, b:JSONObject): DiffEntry[] =>
 
 // combining a list of paths back into a json object (which we can then stringify to put into clipboard):
 // path[] -> Record<string, JSONValue>[] -> JSONObject
-export const toJson = (response) => (diffs: JSONObject[]) =>
+export const toJson = (response: unknown) => (diffs: DiffEntry[]) =>
   R.pipe(
-    R.map(({ keyPath, keyPathString }) => ({ [keyPathString]: R.path(keyPath as string[], response) })),
+    R.map<DiffEntry, JSONObject>(({ keyPath, keyPathString }) => ({ [keyPathString]: R.path(keyPath as string[], response) ?? null })),
```
arseny-kostenko: is `?? null` to satisfy TS? because `path` can be value | undefined, so this is an explicit conversion of undefined to null. There is also `R.pathOr` which can be then `R.pathOr(null, keyPath...`

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,334 @@
...
+
+export async function reconcileClusterMetricsServers(
+  clusterNodesRegistry: ClusterRegistry, 
+  metricsServerMap: MetricsServerMap, 
+  connectionDetails: ConnectionDetails, 
+) {
+  let clusterIds = Object.keys(clusterNodesRegistry) 
+  if (clusterIds.length === 0) {
+    try {
+      const client = await connectToInitialValkeyNode(connectionDetails)
```
arseny-kostenko: It seems reconcileClusterMetricsServers runs in a loop and calls connectToInitialValkeyNode which calls await GlideClient.createClient every time, so it creates a new client every time, and I don't see it closing the client, i.e. this implementation is leaking clients. So either maintain a cluster-client map of clients or in your runReconcileLoop — add finally block where you explicitly close the client, which requires either maintaining that map anyway or making reconcileClusterMetricsServers return that client.

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,334 @@
...
+      res.status(404).send("Invalid nodeId")
+    }
+  })
+
+  router.post("/ping", async (req: Request, res: Response) => {
+    const { nodeId } = req.body
+    const entry = metricsServerMap.get(nodeId)
+    if (entry) {
+      entry.lastSeen = Date.now().toString()
+      res.send("Health ping received for node")
```
arseny-kostenko: let's just res.sendStatus(200) without the body to reduce traffic noise

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,334 @@
...
+      }
+      console.error(err)
+    }
+  }
+  return clusterNodesRegistry
+}
+
+async function findDiff(metricsServerMap: MetricsServerMap, clusterNodeMap: ClusterNodeMap) {
+  // These are nodes that are in the clusterMap but not metricsMap
+  const nodesToAdd: ClusterNodeMap = Object.fromEntries(
```
arseny-kostenko, line 169: this might be simpler with [pickBy](https://ramdajs.com/docs/#pickBy)?
```
const nodesToAdd = R.pickBy(
  (val, key) => !metricsServerMap.has(key),
  clusterNodeMap,
)
```

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,334 @@
...
+    else {
+      res.status(404).send("Invalid nodeId")
+    }
+  })
+
+  router.post("/ping", async (req: Request, res: Response) => {
+    const { nodeId } = req.body
+    const entry = metricsServerMap.get(nodeId)
+    if (entry) {
+      entry.lastSeen = Date.now().toString()
```
arseny-kostenko: let's store a raw: number timestamp — much easier to do all the TTL math against a number of milliseconds instead of converting the string to date

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,334 @@
...
+  // These are nodes that are in the clusterMap but not metricsMap
+  const nodesToAdd: ClusterNodeMap = Object.fromEntries(
+    Object.entries(clusterNodeMap)
+      .filter(([key]) => !metricsServerMap.has(key)),
+  )
+  const now = Date.now()
+  // These are nodes that are in the metricsMap but not in clusterMap or stale nodes
+  const nodesToRemove: string[] = Array.from(metricsServerMap.entries())
+    .filter(([key, value]) => {
+      return !clusterNodeMap[key] || now - Number(value.lastSeen) > ttl
```
arseny-kostenko: once `lastSeen` switched to number — no need for this Number(..)

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Great catch. I'll fix it.

common/src/json-utils.ts
ravjotbrar: The changes in this file were purely to satisfy TS. Changing to use pathOr

apps/metrics/src/index.js
ravjotbrar: Done

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Results in type errors that I don't want to get into at the moment. I could add a todo for the future?

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Sounds good.

apps/server/src/metrics-orchestrator.ts
arseny-kostenko: sure

apps/server/src/metrics-orchestrator.ts
ravjotbrar: This was how I chose to fix it:
```
let initialClient: GlideClient | null = null

export async function getInitialClient() {
  if (!initialClient) {
    initialClient = await connectToInitialValkeyNode(initialConnectionDetails)
  }
  return initialClient
}
```

apps/server/src/__tests__/connection.test.ts
```diff
@@ -2,12 +2,13 @@
...
 import assert from "node:assert"
 import { GlideClient, GlideClusterClient } from "@valkey/valkey-glide"
-import { sanitizeUrl } from "common/src/url-utils.ts"
-import { connectToValkey, isDuplicateConnection } from "../connection.ts"
-import { resolveHostnameOrIpAddress, dns } from "../utils.ts"
-import { checkJsonModuleAvailability } from "../check-json-module.ts"
-import { KEY_EVICTION_POLICY, VALKEY } from "../../../../common/src/constants.ts"
-import { ConnectionDetails } from "../actions/connection.ts"
+import { sanitizeUrl } from "valkey-common"
+import { KEY_EVICTION_POLICY, VALKEY } from "valkey-common"
```
arseny-kostenko: import from the same module as above

apps/server/src/__tests__/metrics-orchestrator.test.ts
```diff
@@ -0,0 +1,205 @@
...
+import assert from "node:assert"
+import { GlideClient } from "@valkey/valkey-glide"
+import {
+  metricsServerMap,
+  stopAllMetricsServers,
+  reconcileClusterMetricsServers,
+  clients,
+  __test__ 
+} from "../metrics-orchestrator"
+import type { ClusterNodeMap, MetricsServerMap } from "../metrics-orchestrator"
```
arseny-kostenko: import from the same module as above

apps/server/src/connection.ts
```diff
@@ -1,12 +1,13 @@
 import { GlideClient, GlideClusterClient, InfoOptions, ServerCredentials } from "@valkey/valkey-glide"
 import * as R from "ramda"
 import WebSocket from "ws"
-import { VALKEY } from "../../../common/src/constants"
+import { VALKEY } from "valkey-common"
+import { sanitizeUrl } from "valkey-common"
+import { type KeyEvictionPolicy } from "valkey-common"
```
arseny-kostenko: imports

apps/server/tsconfig.json
```diff
@@ -36,5 +29,8 @@
   "exclude": [
     "node_modules",
     "dist"
-  ]
+  ],
+  // "references": [
```
arseny-kostenko: need this?

apps/server/src/metrics-orchestrator.ts
```diff
@@ -0,0 +1,349 @@
...
+    env: {
+      ...process.env,
+      PORT: "0",
+      VALKEY_HOST: nodeToStart.host,
+      VALKEY_PORT: String(nodeToStart.port),
+      VALKEY_USERNAME: nodeToStart.username,
+      VALKEY_PASSWORD: nodeToStart.password,
+      VALKEY_TLS: String(nodeToStart.tls),
+      VALKEY_VERIFY_CERT: String(nodeToStart.verifyTlsCertificate),
+      DATA_DIR: `${data_dir}/${nodeId}`,
```
ArgusLi, line 245: If data_dir is not set, metrics creates an "undefined" directory. We should have a fallback e.g. `process.env.DATA_DIR ?? "./data"`

PR #245, opened 2026-03-20
Make server host and port configurable through env variables

ravjotbrar:
## Description

This PR makes the Valkey Admin server host and port configurable through environment variables, improving flexibility for deployment across different environments.

It also makes MAX_CONNECTIONS configurable via MAX_CONNECTIONS environment variable
Defaults to Infinity when env var is not set (instead of hardcoded 12).

Finally, it improves the HotKeys tooltip description for clarity.

### DIFF
--- a/apps/frontend/src/state/epics/wsEpics.ts
+++ b/apps/frontend/src/state/epics/wsEpics.ts
@@ -28,15 +28,11 @@ let socket$: WebSocketSubject<PayloadAction> | null = null
 
 const isBrowser = typeof window !== "undefined"
 
-const backendHost =
-  process.env.SERVER_HOST ||
-  (isBrowser ? window.location.hostname : null) ||
-  "localhost"
-
-const backendPort =
-  process.env.SERVER_PORT ||
-  (isBrowser ? window.location.port : null) ||
-  "8080"
+const url =
+  process.env.WS_URL ||
+  (isBrowser
+    ? `ws://${window.location.hostname}:${window.location.port || "8080"}`
+    : "ws://localhost:8080")
 
 const connect = (store: Store) =>
   action$.pipe(
@@ -48,7 +44,7 @@ const connect = (store: Store) =>
       }
       const createSocket = () => {
         socket$ = webSocket({
-          url: `ws://${backendHost}:${backendPort}`,
+          url,
           deserializer: (message) => JSON.parse(message.data),
           serializer: (message) => JSON.stringify(message),
           openObserver: {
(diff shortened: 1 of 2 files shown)


apps/frontend/src/state/epics/wsEpics.ts
```diff
@@ -26,6 +26,18 @@ import {
 
 let socket$: WebSocketSubject<PayloadAction> | null = null
 
+const isBrowser = typeof window !== "undefined"
```
ArgusLi, line 29: Trying to understand the mechanism of window. Where is it coming from?

Is the idea that if it's run as part of an electron app, window will be automatically populated?

apps/metrics/src/index.js
```diff
@@ -143,8 +143,17 @@ async function main() {
 
   // Setting port to 0 means Express will dynamically find a port
   const port = Number(cfg.server.port || 0)
-  const backendServerHost = process.env.SERVER_HOST ?? "localhost"
-  const backendServerPort = process.env.SERVER_PORT ?? "8080"
+  const isBrowser = typeof window !== "undefined"
```
ArgusLi: Same question here, and this is in the metrics app instead of the frontend app.

ravjotbrar: This is a mistake, will fix.

apps/frontend/src/state/epics/wsEpics.ts
ravjotbrar: The idea is if it's run as a web app, it'll be automatically populated. If its electron it should pass through to the default options of localhost:8080.

apps/metrics/src/index.js
```diff
@@ -143,8 +143,10 @@ async function main() {
...
+
+  const backendServerHost =process.env.SERVER_HOST || "localhost"
```
ArgusLi: Nit: space missing on the other half of the =

apps/frontend/src/state/epics/wsEpics.ts
```diff
@@ -36,7 +48,7 @@ const connect = (store: Store) =>
       }
       const createSocket = () => {
         socket$ = webSocket({
-          url: "ws://localhost:8080",
+          url: `ws://${backendHost}:${backendPort}`,
```
arseny-kostenko: it can be `ws` or `wss` depending on the origin, so need to calculate that too. Probably easier to just have a single WS_URL env variable that includes the protocol, host, port and path instead of all these calculations

ravjotbrar: Good point

PR #246, opened 2026-03-23
added error boundaries to protect Fronted  against server crashes

nassery318:
## Description

Adding error boundaries to protect the FE against the server crashes

1. Added Try Catch blocks
2. Null assertions fixed
3. Handling missing Metrics URI
4. And timeout for fetch requests to avoid indefinite waiting

### Change Visualization

### DIFF
--- a/apps/server/src/actions/commandLogs.ts
+++ b/apps/server/src/actions/commandLogs.ts
@@ -80,9 +80,15 @@ export const commandLogsRequested = withDeps<Deps, void>(
     const { connectionId, clusterId } = action.payload
     const connectionIds = clusterId ? clusterNodesMap.get(clusterId as string) ?? [] : [connectionId]
     const commandLogType: CommandLogType = action.payload.commandLogType as CommandLogType
-    
+
     const promises = connectionIds.map(async (connectionId: string) => {
       const metricsServerURI = metricsServerMap.get(connectionId)?.metricsURI
+
+      if (!metricsServerURI) {
+        sendCommandLogsError(ws, connectionId, new Error("Metrics server URI not found"))
+        return
+      }
+
       try {
         const url = `${metricsServerURI}/commandlog?type=${commandLogType}`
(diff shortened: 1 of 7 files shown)


apps/server/src/actions/commandLogs.ts
```diff
@@ -80,9 +80,15 @@ export const commandLogsRequested = withDeps<Deps, void>(
     const { connectionId, clusterId } = action.payload
     const connectionIds = clusterId ? clusterNodesMap.get(clusterId as string) ?? [] : [connectionId]
     const commandLogType: CommandLogType = action.payload.commandLogType as CommandLogType
-    
+
     const promises = connectionIds.map(async (connectionId: string) => {
       const metricsServerURI = metricsServerMap.get(connectionId)?.metricsURI
+
+      if (!metricsServerURI) {
```
arseny-kostenko, line 87: I see this pattern repeated in lots of handlers. Makes me think that a better fix would be one level above at
```
await handler({ ws, clients, connectionId: connectionId!, metricsServerMap, clusterNodesMap })(action as ReduxAction)
```
and instead of `metricsServerMap` — provide `metricsServerURI`, something like
```
const metricsServerURI = metricsServerMap.get(connectionId)?.metricsURI
if (!metricsServerURI) await noMetricsServiceFound({ connectionId })
else await handler({ ws, clients, connectionId: connectionId!, metricsServerURI, clusterNodesMap })(action as ReduxAction)
```
that way, you will make the handlers even smaller

nassery318: This would only work for single-connection handlers. For cluster handlers that iterate over multiple connectionIds, each iteration needs its own metricsServerURI lookup. 

as shown in CommandLogs: 
`const connectionIds = clusterId ? clusterNodesMap.get(clusterId) ?? [] : [connectionId]
  const promises = connectionIds.map(async (connectionId: string) => {
  const metricsServerURI = metricsServerMap.get(connectionId)?.metricsURI`

PR #247, opened 2026-03-23
FEATURE: Add metrics file storage limit

ArgusLi:
## Description

This PR introduces **size-based storage management** for the metrics service's NDJSON data. Previously, metrics data grew without any size constraint (only time-based retention via `retention_days` existed). Now, each metric type is allocated a weighted share of a configurable total storage budget (`retention_size_mb`), and data is written in a **ring-buffer style** — oldest entries are deleted when the per-metric cap is reached.
 
### 1. Size-Aware NDJSON Writer
- Rewrote the writer to produce **sequenced outputs** using a `{prefix}_{YYYYMMDD}_{seq}.ndjson` naming convention.
- Rotation to a new sequence number when `maxFileSize` is exceeded.
- When the number of entries for a prefix reaches `maxFiles`, the **oldest entries are deleted first** (ring-buffer eviction).
- On restart, the writer **discovers the highest existing sequence number** to avoid overwriting data.
- Large batches that exceed a single output's budget are **chunked across multiple outputs** automatically.
 
### 2. Sequenced Reading in the Streamer
- Updated `filesFor()` to parse and sort by `{date}_{seq}` so data is read back in the correct chronological order.
- The `parseFile()` helper extracts date and sequence from the new naming format.
 
### 3. Per-Metric Capacity Allocation in Collector Initialization
- Added `CAPACITY_WEIGHTS` — a weight map giving each metric type a proportional share of total storage (e.g., `memory: 0.30`, `monitor: 0.25`, `cpu: 0.10`, etc.).
- `normalizeWeights()` re-normalizes weights to active epics only, so unused epic budgets are redistributed.
- `computeCapacity()` converts a weight + total MB into concrete `{ maxFiles, maxFileSize }` values, clamped between `MIN_FILE_SIZE` (256 KB) and `MAX_FILE_SIZE` (10 MB) with a minimum of `MIN_FILES` (4).
- `warnIfStorageTooSmall()` logs a warning if `retention_size_mb` is below what's needed to satisfy minimums.
- Both `setupCollectors()` and `startMonitor()` now pass computed `maxFiles`/`maxFileSize` to each writer.
 
### 4. Config Changes
- Added `retention_size_mb: 50` default to the `storage` config section.
- `RETENTION_SIZE` env var override support added.

### DIFF
--- a/apps/metrics/config.yml
+++ b/apps/metrics/config.yml
@@ -9,6 +9,11 @@ collector:
   batch_ms: 60000
   batch_max: 500
 
+# Each epic controls its own storage retention:
+#   data_retention_mb   – max disk space (MB) for this epic's .ndjson files.
+#                         Oldest files are evicted when the budget is exceeded.
+#   data_retention_days – files older than this (by birthtime) are deleted
+#                         during the daily cleanup sweep.
 epics:
   - name: "memory"
     type: "memory_stats"
--- a/apps/metrics/src/effects/ndjson-writer.js
+++ b/apps/metrics/src/effects/ndjson-writer.js
@@ -54,12 +54,15 @@ export const makeNdjsonWriter = ({ dataDir, filePrefix, maxFiles, maxFileSize })
 
   const evictOldest = async () => {
     const files = await listPrefixFiles(dataDir, filePrefix)
-    let toDelete = files.length - maxFiles + 1 // +1 to make room for the new file
-    for (const f of files) {
-      if (toDelete <= 0) break
-      await fs.promises.unlink(path.join(dataDir, f)).catch(() => {})
-      toDelete--
-    }
+    const toDeleteCount = Math.max(0, files.length - maxFiles + 1) // +1 to make room for the new file
+    
+    const toDelete = files.slice(0, toDeleteCount)
+
+    await Promise.all(
+      toDelete.map((f) =>
+        fs.promises.unlink(path.join(dataDir, f)).catch(() => {}),
+      ),
+    )
   }
 
   const advanceSeq = async () => {
--- a/apps/metrics/src/config.js
+++ b/apps/metrics/src/config.js
@@ -10,6 +10,11 @@ const cfgPath = process.env.CONFIG_PATH || path.join(__dirname, "..", "config.ym
 
 let config = null
 
+/**
+ * Per-epic retention defaults, applied to any epic missing these fields.
+ * @property {number} data_retention_mb   – max disk budget (MB) per epic. Oldest files evicted when exceeded.
+ * @property {number} data_retention_days – files older than this (by birthtime) are deleted in the daily cleanup.
+ */
 const EPIC_DEFAULTS = { data_retention_mb: 10, data_retention_days: 30 }
 
 const DEFAULTS = {


apps/metrics/src/effects/ndjson-streamer.js
```diff
@@ -7,9 +7,23 @@ const DATA_DIR = process.env.DATA_DIR || path.resolve(process.cwd(), "data")
 
 const dayStr = (date) => date.toISOString().slice(0, 10).replace(/-/g, "")
 
-const fileFor = (prefix, date) => {
-  const dataDir = DATA_DIR
-  return path.join(dataDir, `${prefix}_${dayStr(date)}.ndjson`)
+const parseFile = (f) => {
```
arseny-kostenko: confusing name — it's parsing the file name, not the file

arseny-kostenko: more specifically, it parses file's date and sequence

arseny-kostenko: any reason for not using mtime for sorting instead of relying on file naming convention?

ArgusLi: I'll change it to use birthtime. I was mostly following the ndjson-streamer's existing strategy. It definitely makes more sense to use statistics for date here.

apps/metrics/src/effects/ndjson-writer.js
```diff
@@ -4,17 +4,89 @@ import * as R from "ramda"
...
+    if (match) maxSeq = Math.max(maxSeq, Number(match[1]))
+  }
+  return maxSeq
+}
+
+const listPrefixFiles = async (dataDir, filePrefix) => {
+  const fileNames = await fs.promises.readdir(dataDir).catch(() => [])
+  return fileNames
+    .filter((f) => f.startsWith(`${filePrefix}_`) && f.endsWith(".ndjson"))
+    .sort()
```
arseny-kostenko: I think, this sort will break because it's lexicographical, i.e. `file-10` will come above `file-2`

ArgusLi: That's true. And we should be getting file stats, for each file, sorting by date, then seq number anyways.

apps/metrics/src/effects/ndjson-writer.js
```diff
@@ -4,17 +4,89 @@ import * as R from "ramda"
...
+    const currentSize = await fs.promises.stat(currentFile).then((s) => s.size).catch(() => 0)
+    if (currentSize >= maxFileSize) {
+      await advanceSeq()
+      return { file: path.join(dataDir, `${filePrefix}_${day}_${seq}.ndjson`), size: 0 }
+    }
+
+    return { file: currentFile, size: currentSize }
+  }
+
+  const rotate = async () => {
```
arseny-kostenko: this doesn't rotate, it only deletes

ArgusLi: Let me rename it. I thought of it as more of a rotation, but I think it makes more sense just moving forward and not truly rotating. The name stuck when it no longer rotates. I'll rename.

apps/metrics/src/init-collectors.js
```diff
@@ -20,6 +21,50 @@ const updateCollectorMeta = (name, patch) => {
   return next
 }
 
+// Metric files capacity weights per metric
+const CAPACITY_WEIGHTS = {
```
arseny-kostenko: this makes it non-configurable for the user (as it's hardcoded). I'd expect this to be in `config.yml`. I understand the sentiment of trying to simplifying the configuration for the user like "just set one number — your storage capacity" but imho this is unnecessary overengineering.

arseny-kostenko: let the user set the values in MB in config.yml, don't calculate arbitrary values

apps/metrics/src/effects/ndjson-writer.js
arseny-kostenko: and rotating is more about moving the pointer of the current file to the new file — deleting an old file is not even a rotation concern

apps/metrics/src/effects/ndjson-writer.js
```diff
@@ -1,20 +1,100 @@
...
+      await advanceSeq()
+      return { file: path.join(dataDir, `${filePrefix}_${day}_${seq}.ndjson`), size: 0 }
+    }
 
-export const makeNdjsonWriter = ({ dataDir, filePrefix }) => {
-  const fileFor = (ts) => path.join(dataDir, `${filePrefix}_${dayStr(ts, filePrefix)}.ndjson`)
+    return { file: currentFile, size: currentSize }
+  }
+
+  const evictOldest = async () => {
```
arseny-kostenko, line 55: I don't think you need a loop and managing the count, can just split (and make it parallel too):
```
const evictOldest = async () => {
  const files = await listPrefixFiles(dataDir, filePrefix)
  const toDeleteCount = Math.max(0, files.length - maxFiles + 1)

  const toDelete = files.slice(0, toDeleteCount)

  await Promise.all(
    toDelete.map((f) =>
      fs.promises.unlink(path.join(dataDir, f)).catch(() => {})
    )
  )
}
```

apps/metrics/config.yml
```diff
@@ -9,43 +9,54 @@ collector:
...
-storage:
-  retention_days: 5
-
 epics:
   - name: "memory"
     type: "memory_stats"
     poll_ms: 5000
     file_prefix: "memory"
+    data_retention_mb: 15
+    data_retention_days: 5
```
arseny-kostenko, line 23: shouldn't it be one XOR the other? in that case, one line should be commented and there must be a fatal error throw at the startup that two values for the same policy are specified

arseny-kostenko: and leave a comment here, in yaml too

apps/metrics/src/effects/ndjson-writer.js
ArgusLi: Much better.

apps/metrics/config.yml
ArgusLi: I don't think there's actually a mutually exclusive relationship between the size limit and date limit. It just deletes whatever hits a limit first. 
I.e. if there's a metric that doesn't produce a lot, there can be a metric file that is old enough to hit date limit.

I'll add some comments to make sure it's clear to the user.

PR #249, opened 2026-03-25
Fix issues with ElastiCache deployments

ravjotbrar:
## Description
This PR fixes issues with CONFIG not being supported in ElastiCache. Additionally, it changes the WS_URL env variable name to VALKEY_ADMIN_WS_URL for clarity. Finally, it fixes an issue with checking if the application is running in the browser by using window.location.protocol.

### DIFF
--- a/apps/server/src/check-json-module.ts
+++ b/apps/server/src/check-json-module.ts
@@ -4,6 +4,7 @@ export async function checkJsonModuleAvailability(
   client: GlideClient | GlideClusterClient,
 ): Promise<boolean> {
   try {
+    // Elasticache restricts MODULE command
     await client.customCommand(["JSON.TYPE", "nonexistent_key"])
     return true
   } catch {
--- a/apps/server/src/connection.ts
+++ b/apps/server/src/connection.ts
@@ -67,7 +67,10 @@ export async function connectToValkey(
         ["CONFIG", "GET", "maxmemory-policy"],
       ) as [{key: string, value: string}]
 
-      keyEvictionPolicy = evictionPolicyResponse[0].value.toLowerCase() as KeyEvictionPolicy
+      keyEvictionPolicy = R.pipe(
+        R.pathOr("noeviction", [0, "value"]),
+        R.toLower,
+      )(evictionPolicyResponse) as KeyEvictionPolicy
     } catch {
       console.warn("Command \"CONFIG\" not available. Trying \"INFO SERVER\" instead")
       const infoResponse = await standaloneClient.info([InfoOptions.Server])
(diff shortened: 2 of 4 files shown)


apps/server/src/check-json-module.ts
```diff
@@ -1,27 +1,13 @@
...
-    return modules.some((module) => {
-      const moduleName = module.find((attr) => attr.key === "name")?.value
-
-      return typeof moduleName === "string" &&
-             (moduleName.toLowerCase().includes("json") ||
-              moduleName.toLowerCase().includes("rejson"))
-    })
-  } catch (err) {
-    console.error("Error checking JSON module availability:", err)
+    await client.customCommand(["JSON.TYPE", "nonexistent_key"])
```
ArgusLi, line 8: Will there be other types of modules we would need to check in the future?

I like the change as it is much simpler, but would the module list be a more future proof system?

apps/server/src/connection.ts
```diff
@@ -61,8 +61,21 @@ export async function connectToValkey(
...
 
-    const evictionPolicyResponse = await standaloneClient.customCommand(["CONFIG", "GET", "maxmemory-policy"]) as [{key: string, value: string}]
-    const keyEvictionPolicy: KeyEvictionPolicy = evictionPolicyResponse[0].value.toLowerCase() as KeyEvictionPolicy
+    let keyEvictionPolicy: KeyEvictionPolicy = "noeviction"
+    try {
+      const evictionPolicyResponse = await standaloneClient.customCommand(
+        ["CONFIG", "GET", "maxmemory-policy"],
+      ) as [{key: string, value: string}]
+
+      keyEvictionPolicy = evictionPolicyResponse[0].value.toLowerCase() as KeyEvictionPolicy
```
arseny-kostenko: the response can be empty, I suggest 
```
const keyEvictionPolicy = R.pipe(
  R.pathOr("noeviction", [0, "value"]),
  R.toLower,
)(evictionPolicyResponse) as KeyEvictionPolicy
```

apps/server/src/index.ts
```diff
@@ -102,8 +106,18 @@ server.listen(port, () => {
   }
 })
 
-wss.on("connection", (ws: WebSocket) => {
+const interval = setInterval(() => {
```
ArgusLi, line 109: Unlikely case, but if the time taken to ping each client > 30s, then due to the synchronous name of setInterval, the server won't actually do anything.

Would it be better to use exhaustMap? exhaustMap discards timer events if the previous cycle is still running.

apps/server/src/check-json-module.ts
ravjotbrar: Possibly, but it's restricted by ElastiCache.

PR #268, opened 2026-04-09
Fix plain text password being saved in web localstorage

ArgusLi:
## Description

- Remove unnecessary connectionDetails sent back from server to frontend.
- Prevent storage of password if secureStorage is unavailable.
- Add password modal if password is undefined.

### Change Visualization

image

### DIFF
--- a/apps/frontend/src/components/ValkeyReconnect.tsx
+++ b/apps/frontend/src/components/ValkeyReconnect.tsx
@@ -144,7 +144,7 @@ export function ValkeyReconnect() {
 
         <PasswordPromptModal
           connectionLabel={connectionLabel}
-          errorMessage={status === ERROR ? errorMessage ?? undefined : undefined}
+          errorMessage={status === ERROR ? errorMessage : undefined}
           isConnecting={status === CONNECTING}
           onClose={() => setShowPasswordPrompt(false)}
           onSubmit={handlePasswordSubmit}
--- a/apps/frontend/src/components/ValkeyReconnect.tsx
+++ b/apps/frontend/src/components/ValkeyReconnect.tsx
@@ -50,9 +50,7 @@ export function ValkeyReconnect() {
 
   const handlePasswordSubmit = async (password: string) => {
     if (!connection) return
-    const encryptedPassword = password.length > 0 && secureStorage.isAvailable()
-      ? await secureStorage.encrypt(password)
-      : password
+    const encryptedPassword = await secureStorage.encryptIfAvailable(password)
     dispatch(connectPending({
       connectionId: id!,
       connectionDetails: { ...connection.connectionDetails, password: encryptedPassword },
--- a/apps/frontend/src/components/ValkeyReconnect.tsx
+++ b/apps/frontend/src/components/ValkeyReconnect.tsx
@@ -1,11 +1,13 @@
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { useDispatch, useSelector } from "react-redux"
 import { useNavigate, useParams } from "react-router"
 import { CONNECTED, CONNECTING, ERROR } from "@common/src/constants"
 import { Loader2, Database, AlertCircle } from "lucide-react"
 import RetryProgress from "./ui/retry-progress"
+import { PasswordPromptModal } from "./ui/password-prompt-modal"
 import type { RootState } from "@/store"
 import { connectPending } from "@/state/valkey-features/connection/connectionSlice"
+import { secureStorage } from "@/utils/secureStorage"
 
 export function ValkeyReconnect() {
   const dispatch = useDispatch()
@@ -17,6 +19,13 @@ export function ValkeyReconnect() {
   )
 
   const { status, errorMessage, reconnect } = connection || {}
+  const [showPasswordPrompt, setShowPasswordPrompt] = useState(false)
+  const needsPassword = connection?.connectionDetails.password === undefined
+
+  // Close password prompt on successful connection
+  useEffect(() => {
+    if (status === CONNECTED) setShowPasswordPrompt(false)
+  }, [status])
 
   useEffect(() => {
     // Redirect to dashboard or previous location on successful connection
@@ -29,12 +38,31 @@ export function ValkeyReconnect() {
 
   const handleManualReconnect = () => {
     if (!connection) return
-    
+    if (needsPassword) {
+      setShowPasswordPrompt(true)
+      return
+    }
     dispatch(connectPending({
       connectionId: id!,
       connectionDetails: connection.connectionDetails,
     }))
   }
+
+  const handlePasswordSubmit = async (password: string) => {
+    if (!connection) return
+    const encryptedPassword = password.length > 0 && secureStorage.isAvailable()
+      ? await secureStorage.encrypt(password)
+      : password
+    dispatch(connectPending({
+      connectionId: id!,
+      connectionDetails: { ...connection.connectionDetails, password: encryptedPassword },
+    }))
+  }
+
+  const connectionLabel = connection
+    ? connection.connectionDetails.alias
+      || `${connection.connectionDetails.host}:${connection.connectionDetails.port}`
+    : ""
 
   const getNextRetrySeconds = () => {
     if (!reconnect?.nextRetryDelay) return 0
@@ -67,7 +95,7 @@ export function ValkeyReconnect() {
               : "Reconnecting to Valkey..."}
           </h1>
           <p className="text-sm text-gray-600 dark:text-gray-400">
-            {connection.connectionDetails.host}:{connection.connectionDetails.port}
+            {connectionLabel}
           </p>
           <p className="text-sm text-gray-600 dark:text-gray-400">
             {isExhausted
@@ -115,6 +143,15 @@ export function ValkeyReconnect() {
             </button>
           </div>
         )}
+
+        <PasswordPromptModal
+          connectionLabel={connectionLabel}
+          errorMessage={status === ERROR ? errorMessage ?? undefined : undefined}
+          isConnecting={status === CONNECTING}
+          onClose={() => setShowPasswordPrompt(false)}
+          onSubmit={handlePasswordSubmit}
+          open={showPasswordPrompt}
+        />
       </div>
     </div>
   )
--- a/apps/frontend/src/components/ValkeyReconnect.tsx
+++ b/apps/frontend/src/components/ValkeyReconnect.tsx
@@ -144,7 +144,7 @@ export function ValkeyReconnect() {
 
         <PasswordPromptModal
           connectionLabel={connectionLabel}
-          errorMessage={status === ERROR ? errorMessage : undefined}
+          errorMessage={errorMessage}
           isConnecting={status === CONNECTING}
           onClose={() => setShowPasswordPrompt(false)}
           onSubmit={handlePasswordSubmit}
--- a/apps/frontend/src/components/cluster-topology/cluster-node.tsx
+++ b/apps/frontend/src/components/cluster-topology/cluster-node.tsx
@@ -1,13 +1,15 @@
+import { useEffect, useState } from "react"
 import { LayoutDashboard, Terminal, PowerIcon, Server, MemoryStick, Users } from "lucide-react"
 import { useNavigate, useParams } from "react-router"
 import { useSelector } from "react-redux"
-import { CONNECTED, MAX_CONNECTIONS } from "@common/src/constants.ts"
+import { CONNECTED, CONNECTING, ERROR, MAX_CONNECTIONS } from "@common/src/constants.ts"
 import { TooltipProvider } from "@radix-ui/react-tooltip"
 import { Badge } from "../ui/badge"
 import { CustomTooltip } from "../ui/tooltip"
 import { Button } from "../ui/button"
 import { Typography } from "../ui/typography"
 import { HighlightSearchMatch } from "../ui/highlight-search-match"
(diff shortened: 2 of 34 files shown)


apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -73,7 +73,9 @@ export function ClusterNode({
           }
           : primary.password && {
             username: primary.username ?? "",
-            password: await secureStorage.encrypt(primary.password),
+            password: primary.password.length > 0 && secureStorage.isAvailable()
```
ravjotbrar: Do we need to have the password modal here if password is required and it's undefined?

apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -59,36 +61,74 @@ export function ClusterNode({
...
-        tls: primary.tls,
-        verifyTlsCertificate: primary.verifyTlsCertificate,
-        ...(primary.caCertPath && {
-          caCertPath: primary.caCertPath,
-        }),
-        endpointType: "node",
-      }
+  // Look up encrypted password from an existing connection in the same cluster.
+  // Available when secureStorage was active during the original connection.
+  const encryptedPassword = useSelector((state: RootState) =>
```
arseny-kostenko: that should be its own selector getEncryptedPassword(connectionId) or something

apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -59,36 +61,74 @@ export function ClusterNode({
...
+    Object.values(state.valkeyConnection?.connections ?? {}).find(
+      (c) => c.connectionDetails?.clusterId === clusterId && c.connectionDetails?.password,
+    )?.connectionDetails?.password,
+  )
+
+  const [showPasswordModal, setShowPasswordModal] = useState(false)
+
+  // Close password modal on successful connection
+  useEffect(() => {
```
arseny-kostenko: you don't need useEffect at all. `connectionStatus` is from selector, so when the value updates, the components re-render

apps/frontend/src/components/ui/password-prompt-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+    onClose()
+  }
+
+  return (
+    <Dialog.Root onOpenChange={handleClose} open={open}>
+      <Dialog.Portal>
+        <Dialog.Overlay className="fixed inset-0 z-30 bg-black/50" />
+        <Dialog.Content asChild>
+          <div className="fixed inset-0 z-40 flex items-center justify-center">
+            <div className="w-full max-w-sm p-6 bg-white dark:bg-tw-dark-primary dark:border-tw-dark-border rounded-lg shadow-lg border">
```
arseny-kostenko: why unnecessary div? you can still put styles into a single div and use `cn`

apps/frontend/src/components/ui/password-prompt-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+              </Dialog.Description>
+
+              {errorMessage && (
+                <Alert className="mt-4" variant="destructive">
+                  <AlertDescription>{errorMessage}</AlertDescription>
+                </Alert>
+              )}
+
+              <form className="space-y-4 mt-4" onSubmit={handleSubmit}>
+                <div>
```
arseny-kostenko: unnecessary wrapping divs; the entire form can be a simple flex col (based on the screenshot) without all these wrappers

apps/frontend/src/components/connection/Connection.tsx
```diff
@@ -34,6 +39,43 @@ export function Connection() {
...
+      ? await secureStorage.encrypt(password)
+      : password
+    dispatch(connectPending({
+      connectionId: passwordPromptConnectionId,
+      connectionDetails: { ...connection.connectionDetails, password: encryptedPassword },
+      preservedHistory: connection.connectionHistory,
+    }))
+  }
+
+  const promptedConnection = passwordPromptConnectionId
```
arseny-kostenko: you don't need a ternary here

apps/frontend/src/components/ValkeyReconnect.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { useNavigate, useParams } from "react-router"
 import { CONNECTED, CONNECTING, ERROR } from "@common/src/constants"
 import { Loader2, Database, AlertCircle } from "lucide-react"
 import RetryProgress from "./ui/retry-progress"
+import { PasswordPromptModal } from "./ui/password-prompt-modal"
 import type { RootState } from "@/store"
 import { connectPending } from "@/state/valkey-features/connection/connectionSlice"
+import { secureStorage } from "@/utils/secureStorage"
 
 export function ValkeyReconnect() {
```
arseny-kostenko, line 12: where is this used?

arseny-kostenko: nvm, found it

apps/frontend/src/components/cluster-topology/cluster-node.tsx
ArgusLi: But re-render doesn't change useState status no? So even after connection success, the modal stays open.

apps/frontend/src/components/ValkeyReconnect.tsx
```diff
@@ -29,13 +38,32 @@ export function ValkeyReconnect() {
...
+    }
     dispatch(connectPending({
       connectionId: id!,
       connectionDetails: connection.connectionDetails,
     }))
   }
 
+  const handlePasswordSubmit = async (password: string) => {
+    if (!connection) return
+    const encryptedPassword = password.length > 0 && secureStorage.isAvailable()
```
arseny-kostenko: there seems to be at least 5 occurrences of this code — time to extract it into a function

apps/frontend/src/components/connection/Connection.tsx
ArgusLi: Leads to a ts error without.

apps/frontend/src/components/ValkeyReconnect.tsx
```diff
@@ -115,6 +143,15 @@ export function ValkeyReconnect() {
             </button>
           </div>
         )}
+
+        <PasswordPromptModal
+          connectionLabel={connectionLabel}
+          errorMessage={status === ERROR ? errorMessage ?? undefined : undefined}
```
arseny-kostenko: `?? undefined` literally changes nothing

arseny-kostenko: can you have circumstances when status !== ERROR && errorMessage != null?

apps/frontend/src/state/valkey-features/connection/connectionSlice.test.ts
```diff
@@ -47,7 +47,7 @@ describe("connectionSlice", () => {
           host: "localhost",
           port: "6379",
           username: "admin",
-          password: "secret",
+          password: undefined,
```
arseny-kostenko, line 50: should we update the it("should...") part too?

apps/frontend/src/components/ui/edit-form.tsx
```diff
@@ -41,7 +41,7 @@ function EditForm({ onClose, connectionId }: EditFormProps) {
     endpointType: "node" as const,
     authType: "password",
   })
-  const [passwordDirty, setPasswordDirty] = useState(false)
+  const [passwordChanged, setPasswordChanged] = useState(false)
```
arseny-kostenko, line 44: any benefit from renaming passwordChanged to passwordDirty?

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -217,6 +223,7 @@ export const autoReconnectEpic = (store: Store) =>
 
       const disconnectedConnections = Object.entries(connections)
         .filter(([, connection]) => connection.status === DISCONNECTED)
+        .filter(([, connection]) => connection.connectionDetails.password !== undefined)
```
arseny-kostenko: there is a bunch of places where `password === ""` — any impact on this line?

apps/frontend/src/components/connection/Connection.tsx
arseny-kostenko: what error?

apps/server/src/connection.ts
```diff
@@ -116,21 +116,21 @@ export async function connectToValkey(
...
     console.log("Connected to standalone")
 
     subscribe(payload.connectionId, ws)
     ws.send(
-      JSON.stringify({
-        type: VALKEY.CONNECTION.standaloneConnectFulfilled,
-        payload: {
-          connectionId,
-          connectionDetails: {
-            ...payload.connectionDetails,
```
ravjotbrar, line 128: Removing connectionDetails entirely might cause bugs. Let's just remove credentials and keep everything else.

Regarding the cluster bug we ran into, we could change setData to derive an address from connection details or we could pass the address here.

apps/frontend/src/components/cluster-topology/cluster-node.tsx
arseny-kostenko: you can do open={showPasswordModal && connectionStatus !== CONNECTED} — this way you also don't let people open connection modal for connected instances

apps/frontend/src/components/connection/Connection.tsx
arseny-kostenko: you can do `const promptedConnection =  connections[passwordPromptConnectionId as string]`

apps/server/src/connection.ts
arseny-kostenko: yeah I was thinking the same. You can use R.omit

apps/frontend/src/components/ValkeyReconnect.tsx
ArgusLi: errorMessage can be null. ?? undefined gets rid of the ts linting.

apps/server/src/connection.ts
ArgusLi: None of the connectionDetails are used apart from these 2. I think it's better to be explicit rather than implicit to avoid sending back too much info to the frontend.

ArgusLi: I'd like to split the cluster bug into a separate ticket.

apps/frontend/src/components/cluster-topology/cluster-node.tsx
ArgusLi: Ah that's right 🤦‍♂️

apps/frontend/src/components/ValkeyReconnect.tsx
arseny-kostenko: so if we already have ConfigState { errorMessage?: string | null } — why are we expanding the union type with undefined? PasswordPromptModalProps defines `errorMessage?: string` — we're bending the contract for no reason. `errorMessage: null` means there is no error; `errorMessage: undefined` means `errorMessage` field doesn't exist.

apps/frontend/src/components/connection/Connection.tsx
ArgusLi: Wouldn't the ternary make it more clear though?

apps/frontend/src/components/ui/edit-form.tsx
ArgusLi: For me passwordChanged is more descriptive than passwordDirty. That's it.

apps/frontend/src/state/epics/valkeyEpics.ts
ArgusLi: Yes you're right. Missed that.

apps/frontend/src/components/ValkeyReconnect.tsx
ArgusLi: So are you saying that errorMessage shouldn't even check for status === ERROR? Essentially just returning errorMessage if it's not null, and leaving it undefined if null. Because there is no way to get an errorMessage if the status is not ERROR right?

ArgusLi: i.e. `errorMessage={errorMessage ?? undefined}`

arseny-kostenko: errorMessage should have an actual message only when status === ERROR, so whether errorMessage is null or undefined — doesn't (shouldn't) matter, and indeed, in your own PasswordPromptModal component you do `errorMessage &&`.

even this `errorMessage={errorMessage ?? undefined}` is excessive and unnecessary

arseny-kostenko: nobody should ever do `?? undefined` under any circumstances, it's plain wrong. Converting a null to undefined makes no sense and just wrong

arseny-kostenko: even `??` operator is called *nullish* coalescing operator — "a logical operator in JavaScript (ES2020) that returns its right-hand side operand when the left-hand side *is null or undefined*"

ArgusLi: Ok I understand. I've changed the type of password-prompt-modal to string | null, and changed the callers.

apps/frontend/src/components/connection/Connection.tsx
arseny-kostenko: no, it makes it tautological: "if not nil then value otherwise nil"

arseny-kostenko (dismissed):

arseny-kostenko (approved):

PR #271, opened 2026-04-10
Fix password prompt conditions

ravjotbrar:
## Description

Prior to this change, if a user (on Electron) connects to their instance that did not require a password initially, they would be prompted for one upon reconnect. Now the empty password is saved in local storage. 

Additionally, when a user connected with IAM, they would be prompted with a password since there was no guard.

### Change Visualization

### DIFF
--- a/apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
+++ b/apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
@@ -96,7 +96,8 @@ export const ClusterConnectionGroup = ({ clusterId, connections, highlight = "",
 
   const handleConnectLatest = () => {
     if (!lastOpenedNode) return
-    if (lastOpenedNode.connection.connectionDetails.password === undefined && onPasswordRequired) {
+    const { password, authType } = lastOpenedNode.connection.connectionDetails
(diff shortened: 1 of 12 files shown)


apps/frontend/src/components/connection/ClusterConnectionGroup.tsx
```diff
@@ -96,7 +96,8 @@ export const ClusterConnectionGroup = ({ clusterId, connections, highlight = "",
 
   const handleConnectLatest = () => {
     if (!lastOpenedNode) return
-    if (lastOpenedNode.connection.connectionDetails.password === undefined && onPasswordRequired) {
+    const { password, authType } = lastOpenedNode.connection.connectionDetails
+    if (authType !== "iam" && password === undefined && onPasswordRequired) {
```
arseny-kostenko: `password === undefined` — I'm really worried about all these conditions when we have `""` and the ghost of `null` lurking. I'd really suggest we create a util function
```
const isBlank = R.either(R.isNil, R.isEmpty)
```
and do `isBlank(password)`

ravjotbrar: Addressed

arseny-kostenko (approved):

PR #279, opened 2026-04-15
 Aggregate hot keys across cluster nodes

ravjotbrar:
### Hot keys fan-out & aggregation
- Hot key requests are fanned out to all nodes in clusterNodesRegistry when a clusterId is present
- Results are aggregated by summing access counts per key, with size and TTL preserved from the first node that has them 
- Hot keys state is keyed by clusterId for cluster connections

### Error handling
- A yellow warning banner is shown when partial data is returned alongside node errors
- Node-level errors are collected per node and returned alongside data in hotKeysFulfilled rather than firing individual hotKeysError messages 
- If all nodes fail or monitor isn't running, a fulfilled response with empty data is still sent so the UI displays the new banner

### Hot key calculation fix
- calculateHotKeysFromMonitor previously counted every command argument as a key. It now only counts args[0] for single-key commands, all args for mget/json.mget, and every other arg for mset

### Cluster topology fix
- Added refreshAllClusterRegistriesLoop to periodically broadcast updated cluster topology to all connected clients
- Simplified reconcileClusterMetricsServers — removed getInitialClient indirection, increased reconcile interval to 10s

### Misc
- Renamed clusterNodesMap → connectedNodesByCluster throughout for clarity
- Added clusterNodesRegistry to the Deps type so all action handlers have access to the full cluster node map

### Change Visualization

### DIFF
--- a/apps/frontend/src/components/monitoring/hot-keys.tsx
+++ b/apps/frontend/src/components/monitoring/hot-keys.tsx
@@ -45,7 +45,8 @@ export function HotKeys({ data, errorMessage, status, monitorRunning, nodeErrors
   }
 
   const nodeErrorsBanner = nodeErrors && nodeErrors.length > 0 && (
-    <div className="m-3 p-3 bg-yellow-50 dark:bg-yellow-900/20 rounded-md border border-yellow-200 dark:border-yellow-700 flex items-start gap-2">
+    <div className="m-3 p-3 bg-yellow-50 dark:bg-yellow-900/20 rounded-md border
+      border-yellow-200 dark:border-yellow-700 flex items-start gap-2">
       <AlertCircle className="w-4 h-4 text-yellow-500 mt-0.5 shrink-0" />
       <div>
         <Typography variant="bodySm">
--- a/apps/frontend/src/components/monitoring/hot-keys.tsx
+++ b/apps/frontend/src/components/monitoring/hot-keys.tsx
@@ -44,146 +44,152 @@ export function HotKeys({ data, errorMessage, status, monitorRunning, nodeErrors
     return <LoadingState message="Loading hot keys..." />
   }
 
+  const nodeErrorsBanner = nodeErrors && nodeErrors.length > 0 && (
+    <div className="m-3 p-3 bg-yellow-50 dark:bg-yellow-900/20 rounded-md border border-yellow-200 dark:border-yellow-700">
+      <div className="flex items-start gap-2">
+        <AlertCircle className="w-4 h-4 text-yellow-500 mt-0.5 shrink-0" />
+        <div>
+          <Typography variant="bodySm">
+            Hot keys data is partial —{" "}
+            {nodeErrors.length} metrics server{nodeErrors.length > 1 ? "s" : ""} failed to respond or are not connected:
+          </Typography>
+          <ul className="mt-1 space-y-0.5">
+            {nodeErrors.map(({ connectionId, error }) => (
+              <li key={connectionId}>
+                <Typography variant="bodySm">
+                  <span className="font-mono">{connectionId}</span>: {error}
+                </Typography>
+              </li>
+            ))}
+          </ul>
+        </div>
+      </div>
+    </div>
+  )
+
   return sortedHotKeys.length > 0 ? (
     <>
-      {nodeErrors && nodeErrors.length > 0 && (
-        <div className="m-3 p-3 bg-yellow-50 dark:bg-yellow-900/20 rounded-md border border-yellow-200 dark:border-yellow-700">
-          <div className="flex items-start gap-2">
-            <AlertCircle className="w-4 h-4 text-yellow-500 mt-0.5 shrink-0" />
-            <div>
-              <Typography variant="bodySm">
-                Hot keys data is partial — {nodeErrors.length} metrics server{nodeErrors.length > 1 ? "s" : ""} failed to respond:
-              </Typography>
-              <ul className="mt-1 space-y-0.5">
-                {nodeErrors.map(({ connectionId, error }) => (
-                  <li key={connectionId}>
-                    <Typography variant="bodySm">
-                      <span className="font-mono">{connectionId}</span>: {error}
-                    </Typography>
-                  </li>
-                ))}
-              </ul>
-            </div>
-          </div>
-        </div>
-      )}
+      {nodeErrorsBanner}
       <TableContainer
         header={
-        <>
-          <StaticTableHeader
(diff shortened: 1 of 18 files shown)


apps/frontend/src/components/monitoring/hot-keys.tsx
```diff
@@ -43,123 +44,152 @@ export function HotKeys({ data, errorMessage, status, monitorRunning, onKeyClick
     return <LoadingState message="Loading hot keys..." />
   }
 
+  const nodeErrorsBanner = nodeErrors && nodeErrors.length > 0 && (
+    <div className="m-3 p-3 bg-yellow-50 dark:bg-yellow-900/20 rounded-md border border-yellow-200 dark:border-yellow-700">
+      <div className="flex items-start gap-2">
```
arseny-kostenko: unnecessary div

apps/frontend/src/components/monitoring/hot-keys.tsx
```diff
@@ -43,123 +44,152 @@ export function HotKeys({ data, errorMessage, status, monitorRunning, onKeyClick
...
+        <AlertCircle className="w-4 h-4 text-yellow-500 mt-0.5 shrink-0" />
+        <div>
```
arseny-kostenko: since this is still a row, not a column, this appears to be another unnecessary div

apps/frontend/src/components/monitoring/hot-keys.tsx
```diff
@@ -43,123 +44,152 @@ export function HotKeys({ data, errorMessage, status, monitorRunning, onKeyClick
...
-                  errorMessage
-                )}
-              </Typography>
+    <>
+      {nodeErrorsBanner}
+      <EmptyState
+        action={
+          (errorMessage || !monitorRunning) && (
+            <div className="mt-2 p-3 bg-red-50 dark:bg-red-900/20 rounded-md">
+              <div className="flex items-start gap-2">
```
arseny-kostenko: unnecessary div

apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -52,14 +52,25 @@ const ACCESS_COMMANDS = [
...
-        args.forEach((key) => {
-          acc[key] = acc[key] ? acc[key] + 1 : 1
-        })
+      const normalizedCmd = cmd.trim().toLowerCase()
+      if (!ACCESS_COMMANDS.includes(normalizedCmd)) return acc
+
+      if (MULTI_KEY_COMMANDS.has(normalizedCmd)) {
+        args.forEach((key) => { acc[key] = (acc[key] ?? 0) + 1 })
+      } else if (INTERLEAVED_KEY_COMMANDS.has(normalizedCmd)) {
+        for (let i = 0; i < args.length; i += 2) {
```
arseny-kostenko: I think it's cognitively easier if we add some semantics via:
```
R.splitEvery(2, args).forEach(([key]) => { acc[key] = (acc[key] ?? 0) + 1 })
```
instead of operating with an index

apps/server/src/actions/hotkeys.ts
```diff
@@ -61,32 +79,59 @@ export const hotKeysRequested = withDeps<Deps, void>(
...
+    }
+
+    const aggregatedHotKeys = R.pipe(
+      R.chain(({ hotKeys }: HotKeysResponse) => hotKeys as unknown as [string, number, number | null, number][]),
+      R.reduce((acc, [key, count, size, ttl]: [string, number, number | null, number]) => ({
+        ...acc,
+        [key]: [key, (acc[key]?.[1] ?? 0) + count, acc[key]?.[2] ?? size, acc[key]?.[3] ?? ttl],
+      }), {} as Record<string, [string, number, number | null, number]>),
+      R.values,
+      R.sort(([, a]: [string, number], [, b]: [string, number]) => b - a),
```
arseny-kostenko: can be just `R.sort(R.descend(R.prop(1)))`

apps/server/src/connection.ts
```diff
@@ -275,11 +279,12 @@ export async function connectToCluster(
       ws.send(
         JSON.stringify({
           type: VALKEY.CLUSTER.addCluster,
-          payload: { clusterId, clusterNodes }, //TODO: strip credentials (this will impact connecting from Cluster Topology)
```
arseny-kostenko, line 278: is this todo not valid anymore?

apps/server/src/index.ts
```diff
@@ -112,11 +106,45 @@ async function runReconcileLoop() {
   }
 }
 
+async function refreshAllClusterRegistries() {
+  await Promise.all(
+    Object.entries(clusterNodesRegistry).map(async ([clusterId]) => {
+      const clientEntry = [...clients.values()].find((e) => e.clusterId === clusterId)
```
arseny-kostenko, line 112: do you really need destructuring here? isn't client.values() already returning an array?

apps/server/src/index.ts
```diff
@@ -112,11 +106,45 @@ async function runReconcileLoop() {
...
+      if (!clientEntry) return
+
+      const updatedNodes = clusterNodesRegistry[clusterId]
+      // Am I being too defensive here?
```
arseny-kostenko, line 116: hard to tell... maybe you are but we need to check this another time

apps/metrics/src/analyzers/calculate-hot-keys.js
ravjotbrar: Agreed.

apps/server/src/actions/hotkeys.ts
ravjotbrar: Good point

apps/server/src/connection.ts
ravjotbrar: Argus did it in his password storage fix

apps/server/src/index.ts
ravjotbrar: clients.values() returns an iterator. We could do Array.from(clients.values()).find(...) if you prefer.

apps/server/src/connection.ts
ArgusLi: Yes, it was removed as part of https://github.com/valkey-io/valkey-admin/pull/268/changes#diff-919a3938dd36642fcfaf0b96a11e3cea582f14b78e4873914dd9651b452d24ebL300. Not sure why it's back.

arseny-kostenko (dismissed):

arseny-kostenko (approved):

PR #280, opened 2026-04-15
Fix password prompt when there is no password

ArgusLi:
## Description

Previously an empty string was not an option, so I used truthy checks to determine if the password modal was needed. [This change allows for empty passwords to be unencrypted](https://github.com/valkey-io/valkey-admin/pull/271) which changes the assumption the password modal appearance was based on. This change fixes the remaining truthy checks.

### DIFF
--- a/apps/frontend/src/state/valkey-features/connection/connectionSelectors.ts
+++ b/apps/frontend/src/state/valkey-features/connection/connectionSelectors.ts
@@ -22,7 +22,7 @@ export const selectEndpointType = (id: string) => (state: RootState) => atId(id,
 
 export const selectEncryptedPassword = (clusterId: string) => (state: RootState) =>
   Object.values(state.valkeyConnection?.connections ?? {}).find(
-    (c) => c.connectionDetails?.clusterId === clusterId && c.connectionDetails?.password,
+    (c) => c.connectionDetails?.clusterId === clusterId && R.isNotNil(c.connectionDetails?.password),
   )?.connectionDetails?.password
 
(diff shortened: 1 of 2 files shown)


apps/frontend/src/state/valkey-features/connection/connectionSelectors.ts
```diff
@@ -18,8 +18,18 @@ export const selectIsAnyConnecting = (state: RootState) =>
   Object.values(selectConnections(state)).some((c) => c.status === CONNECTING)
 export const selectJsonModuleAvailable = (id: string) => (state: RootState) =>
   atId(id, state)?.connectionDetails?.jsonModuleAvailable ?? false
+export const selectEndpointType = (id: string) => (state: RootState) => atId(id, state)?.connectionDetails.endpointType ?? "node"
+
 export const selectEncryptedPassword = (clusterId: string) => (state: RootState) =>
   Object.values(state.valkeyConnection?.connections ?? {}).find(
-    (c) => c.connectionDetails?.clusterId === clusterId && c.connectionDetails?.password,
+    (c) => c.connectionDetails?.clusterId === clusterId && R.isNotNil(c.connectionDetails?.password),
```
ravjotbrar, line 25: If we already check isNotNil here, do we also need to check in cluster-node.tsx?

ArgusLi: I think it's important to be consistent.

ravjotbrar (approved):

PR #282, opened 2026-04-16
Improve hot keys UX when monitor is not running

ravjotbrar:
## Description
- Metrics server now returns existing hot key data (with monitorRunning: false) instead of a 400 error when monitor is off, so users see last known data rather than an empty state

- Added lastCollectedAt timestamp to the hot keys response, derived from the most recent monitor log row's timestamp, displayed above the table as "Hot Keys last collected at:..."

- Added a red warning banner when monitor is not running that includes a "Start Monitoring" button — shown both above the table (when data exists) and in the empty state

- The start monitoring button is suppressed when LFU + cluster slot stats are enabled, since that path doesn't use the monitor

- Metrics server retries registration with the backend every second for up to 30 seconds before shutting down

### DIFF
--- a/apps/frontend/src/components/monitoring/Monitoring.tsx
+++ b/apps/frontend/src/components/monitoring/Monitoring.tsx
@@ -18,7 +18,8 @@ import type { RootState } from "@/store"
 import { commandLogsRequested, selectCommandLogs } from "@/state/valkey-features/commandlogs/commandLogsSlice"
 import { useAppDispatch } from "@/hooks/hooks"
 import {
-  hotKeysRequested, selectHotKeys, selectHotKeysStatus, selectHotKeysError, selectHotKeysNodeErrors
+  hotKeysRequested, selectHotKeys, selectHotKeysStatus, selectHotKeysError,
+  selectHotKeysNodeErrors, selectHotKeysLastCollectedAt
 } from "@/state/valkey-features/hotkeys/hotKeysSlice"
 import { selectMonitorRunning } from "@/state/valkey-features/monitor/monitorSlice"
 import { getKeyTypeRequested } from "@/state/valkey-features/keys/keyBrowserSlice"
@@ -53,6 +54,7 @@ export const Monitoring = () => {
   const hotKeysStatus = useSelector((state: RootState) => selectHotKeysStatus(hotKeysId)(state))
   const hotKeysErrorMessage = useSelector((state: RootState) => selectHotKeysError(hotKeysId)(state))
   const hotKeysNodeErrors = useSelector((state: RootState) => selectHotKeysNodeErrors(hotKeysId)(state))
+  const hotKeysLastCollectedAt = useSelector((state: RootState) => selectHotKeysLastCollectedAt(hotKeysId)(state))
   const monitorRunning = useSelector(selectMonitorRunning(id!))
   const keys: KeyInfo[] = useSelector(selectKeys(id!))
 
@@ -189,6 +191,7 @@ export const Monitoring = () => {
               <HotKeys
                 data={hotKeysData}
                 errorMessage={hotKeysErrorMessage as string | null}
+                lastCollectedAt={hotKeysLastCollectedAt}
                 monitorRunning={monitorRunning}
                 nodeErrors={hotKeysNodeErrors}
                 onKeyClick={handleKeyClick}
--- a/apps/frontend/src/components/monitoring/Monitoring.tsx
+++ b/apps/frontend/src/components/monitoring/Monitoring.tsx
@@ -22,6 +22,7 @@ import {
   selectHotKeysNodeErrors, selectHotKeysLastCollectedAt
 } from "@/state/valkey-features/hotkeys/hotKeysSlice"
 import { selectMonitorRunning } from "@/state/valkey-features/monitor/monitorSlice"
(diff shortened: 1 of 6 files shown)


ArgusLi (dismissed):

apps/frontend/src/components/monitoring/Monitoring.tsx
```diff
@@ -191,10 +196,11 @@ export const Monitoring = () => {
               <HotKeys
                 data={hotKeysData}
                 errorMessage={hotKeysErrorMessage as string | null}
+                lastCollectedAt={hotKeysLastCollectedAt}
                 monitorRunning={monitorRunning}
                 nodeErrors={hotKeysNodeErrors}
                 onKeyClick={handleKeyClick}
-                onStartMonitoring={() => setConfigOpen(true)}
+                onStartMonitoring={useHotSlots ? undefined : () => setConfigOpen(true)}
```
arseny-kostenko, line 203: let's tighten up the types and keep it a function anyway:
```
{useHotSlots ? () => false : () => setConfigOpen(true)}
```

apps/metrics/src/handlers/monitor-handler.js
```diff
@@ -17,11 +17,16 @@ export const useMonitor = async (res, client, nodeId) => {
   const { isRunning, willCompleteAt: checkAt } = getCollectorMeta(MONITOR) 
   try {
     if (!isRunning) {
-      return res.status(400).json({ error: "Monitor is not running" })
```
arseny-kostenko, line 20: there is a lot of duplication, the actual condition-related difference seems to be only in meta. I'd rewrite it like this:
```
const { isRunning, willCompleteAt: checkAt } = getCollectorMeta(MONITOR)

try {
  const shouldReturnMonitorData = !isRunning || Date.now() > checkAt

  if (!shouldReturnMonitorData) {
    return res.json({ checkAt })
  }

  const rows = await Streamer.monitor()
  // ts is in microseconds, so * 1000 to convert ts to milliseconds 
  const lastCollectedAt = rows.at(-1)?.ts ? rows.at(-1).ts * 1000 : null
  const hotKeys = await Promise.resolve(rows)
    .then(calculateHotKeysFromMonitor)
    .then(enrichHotKeys(client))

  const monitorMeta = !isRunning
    ? {
        monitorRunning: false,
        checkAt: null,
        startedAt: null,
      }
    : toResponse(getCollectorMeta(MONITOR))

  return res.json({
    hotKeys,
    nodeId,
    lastCollectedAt,
    ...monitorMeta,
  })
} catch (e) {
  ...
}
```

apps/metrics/src/index.js
```diff
@@ -180,7 +180,18 @@ async function main() {
...
 
     console.debug(`listening on http://${metricsBindHost}:${assignedPort}`)
-    await registerWithServer()
+
+    const registerWithRetry = async () => {
+      const maxAttempts = 30
+      for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+        const success = await registerWithServer()
+        if (success) return
+        if (attempt < maxAttempts) await new Promise((resolve) => setTimeout(resolve, 1000))
```
arseny-kostenko, line 189: instead of a fixed 1 second delay, maybe let's have it linearly increasing? `1000 * attempt`

ravjotbrar: My reasoning for fixed delays is we want the metrics server to start as soon as possible after a client is added to our clients map.

apps/metrics/src/handlers/monitor-handler.js
ravjotbrar: Much cleaner, will make the change.

apps/frontend/src/components/monitoring/Monitoring.tsx
ravjotbrar: onStartMonitoring is typed as`(() => void)` since it's a click handler. Undefined seems simpler here.

apps/metrics/src/handlers/monitor-handler.js
ravjotbrar: Just noticed shouldReturnMonitor data includes `!isRunning`. In that case if` !shouldReturnMonitorData`, we shouldn't be returning checkAt when the monitor isn't running.

arseny-kostenko: yeah... what if we invert it?
```
const { isRunning, willCompleteAt: checkAt } = getCollectorMeta(MONITOR)

try {
  if (isRunning && Date.now() <= checkAt) {
    return res.json({ checkAt })
  }

  const rows = await Streamer.monitor()
  const lastCollectedAt = rows.at(-1)?.ts ? rows.at(-1).ts * 1000 : null
  const hotKeys = await Promise.resolve(rows)
    .then(calculateHotKeysFromMonitor)
    .then(enrichHotKeys(client))

  return res.json({
    hotKeys,
    nodeId,
    lastCollectedAt,
    ...(isRunning
      ? toResponse(getCollectorMeta(MONITOR))
      : {
          monitorRunning: false,
          checkAt: null,
          startedAt: null,
        }),
  })
} catch (e) {
  ...
}
```

ravjotbrar: Did exactly that just now. That works!

arseny-kostenko (approved):

PR #286, opened 2026-04-16
cluster discovery endpoint no longer piggybacking on connectPending

### LINKED ISSUE #283: Refactor discovery endpoint redirect to use new slice
Currently, the discover endpoint repurposes connectPending to redirect the connection to an actual node. We should detect endpointType before dispatching connectPending and if it's cluster-endpoint, we should dispatch a new action called something like discoveryEndpointPending, which handles the discovery of topology and redirect accordingly.

nassery318:
## Description

1. discovery endpoint no longer piggybacking on connectPending
2. It is using its own discoveryEndpointPending action

### Change Visualization

### DIFF
--- a/apps/frontend/src/state/epics/valkeyEpics.ts
+++ b/apps/frontend/src/state/epics/valkeyEpics.ts
@@ -17,9 +17,13 @@ import {
   type ConnectionState,
   closeConnectionFulfilled,
   closeConnectionFailed,
-  closeConnection,
-  configEndpointRedirect
+  closeConnection
 } from "../valkey-features/connection/connectionSlice"
+import {
+  discoveryEndpointPending,
+  discoveryEndpointFulfilled,
+  clearEndpointDiscovery
+} from "../valkey-features/topology/topologySlice"
 import { sendRequested } from "../valkey-features/command/commandSlice"
 import { setData } from "../valkey-features/info/infoSlice"
 import { action$, select } from "../middleware/rxjsMiddleware/rxjsMiddleware.ts"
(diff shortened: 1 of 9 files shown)


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -108,11 +112,35 @@ export const connectionEpic = (store: Store) =>
...
     action$.pipe(
-      select(configEndpointRedirect),
-      tap(({ payload: { fromId, toId, connectionDetails } }) => {
-        store.dispatch(deleteConnection({ connectionId: fromId, silent: true }))
-        store.dispatch(connectPending({ connectionId: toId, connectionDetails }))
+      select(discoveryEndpointFulfilled),
+      tap(({ payload: { discoveryId, clusterNodes } }) => {
+        const state = store.getState()
+        const discovery = state.valkeyTopology?.discoveries?.[discoveryId]
+        if (!discovery) return
```
ravjotbrar, line 128: Should we be returning an error in this case?

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -108,11 +112,35 @@ export const connectionEpic = (store: Store) =>
...
+
+        const firstNode = Object.values(clusterNodes)[0]
+        if (!firstNode) return
```
ravjotbrar, line 131: Error?

apps/frontend/src/state/epics/valkeyEpics.ts
nassery318: good point

apps/frontend/src/state/epics/valkeyEpics.ts
nassery318: yes will need to return error

ravjotbrar (approved):

PR #298, opened 2026-04-20
Fix cluster state on WebSocket reconnect with shared backend

### LINKED ISSUE #294: BUG: Cluster topology disappears when multiple users share the same backend
This still needs investigation and instructions to reproduce, but essentially the cluster topology exists and then upon refresh no longer exists.

ravjotbrar:
## Description

When multiple users share the same backend, a reconnecting user would receive updateClusterInfo instead of addCluster because the global clients map still held the existing  GlideClusterClient. Since updateClusterInfo is a no-op on an empty frontend state, the cluster topology would never populate after reconnect.

In addition, clusterNodesRegistry was not cleaned up properly, this fix addresses that.

Changes:
- returnExistingClusterClient now sends addCluster instead of updateClusterInfo, so a reconnecting client always rebuilds its cluster state from scratch

- teardownConnection now accepts an optional clusterNodesRegistry and deletes the cluster entry when the last client for that cluster is closed — preventing stale entries 
from causing fan-out to disconnected nodes. This cleanup is skipped when USE_CLUSTER_ORCHESTRATOR=true since the registry is managed by the reconcile loop in that mode
- closeConnection and the WebSocket close handler both pass clusterNodesRegistry to teardownConnection

### Change Visualization

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -248,12 +248,12 @@ wss.on("connection", (ws: AliveWebSocket) => {
     console.log("Client disconnected. Reason:", code, reason.toString())
 
     for (const connectionId of removedIds) {
-      if (getWatcherCount(connectionId) === 0) teardownConnection(connectionId, clients, metricsServerMap)
+      if (getWatcherCount(connectionId) === 0) teardownConnection(connectionId, clients, metricsServerMap, clusterNodesRegistry)
     }
 
     // Clean up any side-entries (e.g., node entries from config endpoint connections)
     for (const [id] of clients) {
-      if (getWatcherCount(id) === 0) teardownConnection(id, clients, metricsServerMap)
+      if (getWatcherCount(id) === 0) teardownConnection(id, clients, metricsServerMap, clusterNodesRegistry)
     }
   })
 })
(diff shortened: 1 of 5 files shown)


apps/server/src/index.ts
```diff
@@ -248,12 +248,12 @@ wss.on("connection", (ws: AliveWebSocket) => {
     console.log("Client disconnected. Reason:", code, reason.toString())
 
     for (const connectionId of removedIds) {
-      if (getWatcherCount(connectionId) === 0) teardownConnection(connectionId, clients, metricsServerMap)
+      if (getWatcherCount(connectionId) === 0) teardownConnection(connectionId, clients, metricsServerMap, clusterNodesRegistry)
```
arseny-kostenko: I think that's where we need to add some grace period of 10-30 seconds before teardownConnection, i.e.
```
setTimeout(() => {
    if (getWatcherCount(connectionId) === 0) teardownConnection(connectionId, clients, metricsServerMap, clusterNodesRegistry)
}, TIMEOUT_BEFORE_TEARDOWN)
```

arseny-kostenko (approved):

PR #302, opened 2026-04-22
revamp the hotkeys view by adding search, filtering, access count fil…

### LINKED ISSUE #285: Improve UX of hotkeys view
### 1. Limited visibility due to pagination/layout
When dealing with hundreds of keys, displaying only four keys per screen is too restrictive and makes it difficult to get a meaningful overview. The layout should support higher information density or more flexible pagination.

### 2. Lack of filtering and search capabilities
There is currently no way to filter or search keys by node and/or key name. This makes it difficult to quickly locate specific data in large datasets.

### 3. Access count lacks context
The `accessCount` metric is not very useful without an associated timespan. It should clearly indicate the duration of the monitoring interval to provide meaningful context.

### 4. No visual grouping by node
Keys are not visually grouped by node, making it harder to understand distribution and relationships. Introducing grouping would improve readability and analysis.

### 5. Insufficient visibility into node-level load distribution
While the cluster-level hot keys leaderboard is valuable, it does not reveal whether load is evenly distributed across nodes.

A visualization is needed to highlight node-level activity. For example, a grid of nodes represented as small tiles with a “heat” indicator (similar to a contribution heatmap) to quickly identify hot or idle nodes.

### 6. No control over access count range
Users cannot easily understand or refine the range of `accessCount` values being displayed.

For example, if 100 keys are shown with counts ranging from 4 to 123, there should be:
- A visible min–max range indicator  
- A filter to adjust the lower bound (e.g., narrowing the view to 100–123)

### 7. Warning banner scalability issue
The warning banner indicating that metrics servers are not running on some nodes does not scale well for large clusters. It should have a fixed size and avoid expanding excessively as the number of affected nodes grows.

nassery318:
…ters and heatmap

## Description

Revamped the Hot Keys view for cluster and standalone nodes with the following improvements 
1. Search and filtering of hotkeys
2. Access count min and max filter and custom ranges
3. Cluster node filter with search capability 
4. Cluster heat map to show nodes with hight concentration of hotkeys 

### Change Visualization

https://github.com/user-attachments/assets/65ef6ef4-35f3-4e57-8738-8671b0c07cd9


apps/frontend/src/components/monitoring/hotkeys/hot-keys-heatmap.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+function snapToBucket(ratio: number): number {
+  return Math.round(ratio * 4) / 4
+}
+
+export function HotKeysHeatmapModal({ open, onClose, data }: HotKeysHeatmapModalProps) {
+  const [hovered, setHovered] = useState<HoveredTile | null>(null)
+  const [selectedBuckets, setSelectedBuckets] = useState<Set<number>>(new Set())
+
+  const nodeStats = data.reduce((acc, [, accessCount, , , nodeId]) => {
```
arseny-kostenko: I don't think we have compiler enabled, so worth wrapping this with useMemo so it's not calculated on every render

apps/frontend/src/components/monitoring/hotkeys/hot-keys-heatmap.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const key = nodeId ?? "Unknown"
+    acc[key] ??= { nodeId: key, count: 0, totalAccess: 0 }
+    acc[key].count += 1
+    acc[key].totalAccess += accessCount
+    return acc
+  }, {} as Record<string, NodeStat>)
+
+  const sorted: NodeStat[] = Object.values(nodeStats).sort((a, b) => b.count - a.count)
+  const max = sorted[0]?.count ?? 1
+  const min = sorted[sorted.length - 1]?.count ?? 0
```
arseny-kostenko: nit: more readable: `sorted.at(-1)?.count ?? 0`

apps/frontend/src/components/monitoring/hotkeys/hot-keys.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  return (
+    <>
+      {banners}
+      <HotKeysHeatmapModal
+        data={sorted}
+        onClose={() => setIsHeatmapOpen(false)}
+        open={isHeatmapOpen}
+      />
+      <HotKeysToolbar
```
arseny-kostenko, line 118: with the number of props in these components it seems like Context would be a good fit, wdyt?

arseny-kostenko (approved):

PR #305, opened 2026-04-23
Ensure metrics server starts for standalone

ravjotbrar:
## Description

Starting metrics server unless it's K8s, so that standalone connections also get a metrics server on web. Also added deployment mode env variable to npm run dev.

### Change Visualization


nassery318 (approved):
LGTM

PR #308, opened 2026-04-23
addressed ui/ux issues in activity/monitoring view and cluster dropdown

### LINKED ISSUE #295: BUG: Trim whitespace from username
Currently, whitespaces are not trimmed in the connection modal. If a user accidentally enters a space before or after their username/password, the connection will fail.

nassery318:
## Description

1. Trim whitespace from connection fields (in connection and edit connection forms)
2. Moved the warning to avoid obstructing buttons (user can minimize it)
3. Added stop all button and functionality to the monitor warning banner
4. Start Monitoring button renamed to Start MONITOR 
5. Changed from Monitoring view to Activity in the App Header/Sidebar
6. Cluster dropdown fixed with scrolling ability and fixed search input

### Change Visualization

https://github.com/user-attachments/assets/94d9d80b-2892-4415-b317-bfd301b44573


apps/frontend/src/components/ui/monitor-warning-banner.tsx
```diff
@@ -65,6 +65,18 @@ export function MonitorWarningBanner() {
...
+              {runningConnections.length > 1 && (
+                <span
+                  className="text-destructive ml-2 underline cursor-pointer hover:text-destructive/80"
+                  onClick={() => {
+                    runningConnections.forEach(({ connectionId }) => {
+                      handleStop(connectionId)
+                    })
+                  }}
+                >
+                  Stop All
```
ravjotbrar: Let's stop monitoring for all nodes the same way we start monitoring for all nodes, in the backend. The frontend may not always be in sync with the current metrics servers running

arseny-kostenko: Agree. And let's remove the ability to stop individual nodes. All or nothing

ArgusLi:
This will close https://github.com/valkey-io/valkey-admin/issues/295

apps/frontend/src/components/ui/monitor-warning-banner.tsx
```diff
@@ -33,11 +33,21 @@ export function MonitorWarningBanner() {
...
   if (runningConnections.length === 0) return null
 
-  const handleStop = (connectionId: string) => {
-    dispatch(saveMonitorSettingsRequested({
-      connectionId,
-      monitorAction: MONITOR_ACTION.STOP,
-    }))
+  const handleStopAll = () => {
+    if (clusterId) {
+      dispatch(saveMonitorSettingsRequested({
```
ravjotbrar: Are we able to call monitorRequested directly? And why do we need an else block?

nassery318: Addressed

apps/frontend/src/components/ui/monitor-warning-banner.tsx
nassery318: addressed

apps/frontend/src/components/ui/monitor-warning-banner.tsx
ravjotbrar: Can we not call monitorRequested instead of saveMonitorSettings?

apps/server/src/actions/monitorAction.ts
```diff
@@ -47,9 +47,24 @@ const sendMonitorError = (
...
+    } else if (clusterId) {
+      // for cluster nodes: fan out to all nodes in the cluster registry that have an active metrics server.
+      connectionIds = Object.keys(clusterNodesRegistry[clusterId as string] ?? {}).filter((id) => metricsServerMap.has(id))
+      resolvedAction = monitorAction as MonitorAction
+    } else {
+      // Standalone start/status: act on the single specified connection.
+      // Standalone stop is handled by stop_all above.
+      connectionIds = [connectionId]
+      resolvedAction = monitorAction as MonitorAction
+    }
```
ravjotbrar: We shouldn't need a new monitor action. Stop should be called already for each node belonging to that cluster.

ravjotbrar: You could probably remove all of the changes here and just dispatch stop action from the frontend. Let me know if that works.

arseny-kostenko (approved):

PR #309, opened 2026-04-23
Update dashboard data

ravjotbrar:
## Description

Prior to this change, dashboard data was only queried and stored once and we would display stale data. Now we dispatch updateData which piggybacks the setData epic.

### DIFF
--- a/apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
+++ b/apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
@@ -1,4 +1,4 @@
-import { createSlice } from "@reduxjs/toolkit"
+import { createAction, createSlice } from "@reduxjs/toolkit"
 import * as R from "ramda"
 
 export interface ReplicaNode {
@@ -45,6 +45,8 @@ interface ClusterState {
   };
 }
 const initialClusterState: ClusterState = {}
+
+export const updateClusterData = createAction<{connectionId: string, clusterId: string}>("updateClusterData")
 
 const clusterSlice = createSlice({
   name: "valkeyCluster",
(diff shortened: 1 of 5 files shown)


apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
```diff
@@ -120,6 +120,10 @@ const clusterSlice = createSlice({
       }
       state.clusters[clusterId].searchableText = searchableText
     },
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
arseny-kostenko: instead of this, what you need to do is creating an action manually
```
const updateData = createAction("updateData")
```

apps/frontend/src/state/valkey-features/info/infoSlice.ts
```diff
@@ -128,6 +128,10 @@ const infoSlice = createSlice({
   name: VALKEY.STATS.name,
   initialState,
   reducers: {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    updateData: (_state, _action) => {
```
arseny-kostenko: as above

apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
ravjotbrar: Done

apps/frontend/src/state/valkey-features/info/infoSlice.ts
ravjotbrar: Done

arseny-kostenko (approved):

PR #318, opened 2026-04-28
Add monitoring documentation, troubleshooting, and fix hotkeys bugs

ravjotbrar:
## Description
- Updated README with chot keys detection documentation andimportant notes
- Added TROUBLESHOOTING.md with known issues and workarounds

### Bugs Fixed

Bug: getClusterSlotStatsEnabled was always returning true
- Fix: getClusterSlotStatsEnabled to check for cpu-usec in response instead of just command existence

Bug: User has no way to control the number of hot keys displayed
- Fix: Added HOT_KEYS_COUNT env var to control max hot keys returned (default 50)

Bug: Monitor had no limit to the number of hot keys displayed
- Fix: Applied hot keys limit to both monitor-based and hot slots aggregation paths

Monitor client had no error handling:
- Fix: monitor-based metrics server crash under heavy load by adding error handlers on ioredis monitor 
client

Key details not showing when selecting hot key
- Fix: hot key click crash when key wasn't in key browser state
- Fix: key details not showing for hot keys by initializing key browser state on demand


apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -55,7 +55,7 @@ const CUT_OFF_FREQUENCY = 1
 const MULTI_KEY_COMMANDS = new Set(["mget", "json.mget"])
 const INTERLEAVED_KEY_COMMANDS = new Set(["mset"])
 
-export const calculateHotKeysFromMonitor = (rows) =>
+export const calculateHotKeysFromMonitor = (rows, limit = 50) =>
```
arseny-kostenko: Some advanced programming advice: the order of arguments should go from most generic/stable (configuration) to most varying (data) — this allows greater functional composition and reuse of the code. 

This can be worth an entire tech talk (there is at least [one](youtube.com/watch?v=m3svKOdZijA)). So to keep this comment practical, consider that CUT_OFF_FREQUENCY is also coming from API or some other place so that it can't be just injected like here. 

So then what? You'll make it `(rows, limit = 50, cutoff = CUT_OFF_FREQUENCY)`? This already looks awkward to say the least. A good practice is to turn this function into a two arguments function where one argument is data, another argument is a configuration object `{ cutoff, limit }`.

Then what if we need to reuse this function with different configurations, e.g. "top 100 hottest keys" (so we need no cutoff but limit = 100), "all keys accessed more than 1000 times" (so we need no limit but cutoff = 1000). Those functions fully reuse calculateHotKeysFromMonitor with different config, and instead of propagating these numbers in a magic-number fashion, you could give them an actual semantic meaning and literally create functions like `topNHottestKeys` and `keysAccessedMoreThanN`, and define them as:
```
const topNHottestKeys = (rows, limit) => calculateHotKeysFromMonitor(rows, { cutoff: 1, limit })
const keysAccessedMoreThanN = (rows, cutoff) => calculateHotKeysFromMonitor(rows, { cutoff, limit: Number.POSITIVE_INFINITY })
// and then use it like
const hotKeys = await Promise.resolve(rows).then((r) => topNHottestKeys(r)).then(enrichHotKeys(client))
```
and then have more meaningful, easier to follow, semantically obvious unit tests instead of tossing magic numbers around and leaving comments explaining their meaning.

But this can be done even better if we apply the data-last principle + partial application:
```
const calculateHotKeysFromMonitor = (config = {}) => rows => ....

const topNHottestKeys = (limit) => calculateHotKeysFromMonitor({ cutoff: 1, limit })
const keysAccessedMoreThanN = (cutoff) => calculateHotKeysFromMonitor({ cutoff, limit: Number.POSITIVE_INFINITY })
// and then use it like
const hotKeys = await Promise.resolve(rows).then(topNHottestKeys).then(enrichHotKeys(client))

// or if you want to keep it inline
const hotKeys = await Promise.resolve(rows).then(calculateHotKeysFromMonitor({ limit, cutoff })).then(enrichHotKeys(client))
```

But what if you need to use it in an imperative flow and you want to avoid awkward currying like below?
```
const rows = await getRows()
const hotKeys = calculateHotKeysFromMonitor({ limit, cutoff })(rows)
``` 

For that, you wrap the variadic function in [R.curry](https://ramdajs.com/docs/#curry):
```
const calculateHotKeysFromMonitor = R.curry((config = {}, rows) => .... )
```

This allows you to call it both as:
```
calculateHotKeysFromMonitor(config)(data)
calculateHotKeysFromMonitor(config, data)
```

apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -75,6 +75,7 @@ export const calculateHotKeysFromMonitor = (rows) =>
     R.toPairs,
     R.sort(R.descend(R.last)),
     R.reject(([, count]) => count <= CUT_OFF_FREQUENCY),
```
arseny-kostenko: while we're at it — can you please make CUT_OFF_FREQUENCY a configurable env variable defaulting to 100. Ideally, I would love to expose it as a config — the same way as we do `monitoringInterval` and `maxCommandsPerRun`. 

btw, we're not exposing `maxCommandsPerRun` in config — it's an issue

ravjotbrar: maxCommandsPerRun is exposed via config file if the user wants to change it there. Do you want it exposed as an env variable?

arseny-kostenko: I want to expose it via UI

arseny-kostenko (approved):

PR #319, opened 2026-04-28
fixed the cluster and node connect button states

### LINKED ISSUE #317: BUG: Add connection state to all connect buttons
Currently, if a user tries to connect using the cluster Connect button or the power icons in cluster topology, there's no connection state. If a user tries to connect with these buttons, they are not given any feedback if the connection is still connecting or if there is an error.

nassery318:
## Description

1. Cluster connect button state fixed and shows Spinner when connecting 
2. Node connect button state fixed and shows Spinner when connecting 
3. Both above cases will show a toast when there is a connection error

### Change Visualization

https://github.com/user-attachments/assets/4df48ade-40a1-4b52-8b26-d6286e5a8d90

### DIFF
--- a/apps/frontend/src/components/cluster-topology/cluster-node.tsx
+++ b/apps/frontend/src/components/cluster-topology/cluster-node.tsx
@@ -119,6 +119,12 @@ export function ClusterNode({
     </div>
   )
 
+  const powerIconTooltip = isConnected   ? "Connected"
+    : isConnecting ? "Connecting..."
+      : isError      ? "Connection failed — click to retry"
+        : isDisabled   ? `Max connections of ${MAX_CONNECTIONS} reached`
+          :                "Not Connected"
+
   return (
     <div className="w-full">
       <TooltipProvider>
@@ -168,7 +174,7 @@ export function ClusterNode({
 
             {/* Actions */}
             <div className="flex items-center gap-2 shrink-0">
-              <CustomTooltip content={isConnected ? "Connected" : isConnecting ? "Connecting..." : isError ? "Connection failed — click to retry" : isDisabled ? `Max connections of ${MAX_CONNECTIONS} reached` : "Not Connected"}>
+              <CustomTooltip content={powerIconTooltip}>
                 {isConnecting ? (
                   <Loader2
                     className="animate-spin text-gray-500"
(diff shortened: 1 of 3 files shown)


apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -166,17 +168,25 @@ export function ClusterNode({
...
-                  className={cn(
-                    "rounded-full p-0.5",
-                    isConnected && "text-green-500 bg-green-100",
-                    !isConnected && isDisabled && "text-gray-300 cursor-not-allowed bg-gray-100",
-                    !isConnected && !isDisabled && "text-gray-400 cursor-pointer bg-gray-100 hover:text-gray-600",
-                  )}
-                  onClick={isDisabled ? undefined : handleNodeConnect}
-                  size={18}
-                />
+              <CustomTooltip content={isConnected ? "Connected" : isConnecting ? "Connecting..." : isError ? "Connection failed — click to retry" : isDisabled ? `Max connections of ${MAX_CONNECTIONS} reached` : "Not Connected"}>
```
arseny-kostenko: that's a long string, can you extract and/or format it to fit 120 chars?

apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -166,17 +168,25 @@ export function ClusterNode({
...
+                {isConnecting ? (
+                  <Loader2
+                    className="animate-spin text-gray-500"
+                    size={21}
+                  />
+                ) : (
+                  <PowerIcon
+                    className={cn(
+                      "rounded-full p-0.5",
+                      isConnected && "text-green-500 bg-green-100",
```
arseny-kostenko: q: does this (and below) look find in dark mode?

apps/frontend/src/components/cluster-topology/cluster-node.tsx
```diff
@@ -166,17 +168,25 @@ export function ClusterNode({
...
+                      isError && "text-red-500 bg-red-100 cursor-pointer hover:text-red-600",
+                      !isConnected && !isError && isDisabled && "text-gray-300 cursor-not-allowed bg-gray-100",
+                      !isConnected && !isError && !isDisabled && "text-gray-400 cursor-pointer bg-gray-100 hover:text-gray-600",
+                    )}
+                    onClick={isDisabled ? undefined : handleNodeConnect}
```
arseny-kostenko, line 192: let's replace undefined with `() => false` or `R.F` so onClick always accepts a `function` instead of `function?`

arseny-kostenko (approved):

PR #327, opened 2026-04-29
Serialize concurrent connects and reuse committed cluster clients

### LINKED ISSUE #316: BUG: Duplicate connections should not fail
Currently, we reject duplicate connections instead of handling in the backend. We need to send a connectionFulfilled action in response to duplicate connections for both standalone and clusters

ArgusLi:
## Description

Fixes:
1. Concurrent connectors with different connectionIds pointing to the same cluster each independently created a cluster client, leaking orphaned clients.
2. Concurrent connectors with the same connectionId could both reach standalone-discovery + cluster-client creation and race to commit into `clients`.
3. Sequential reconnects with the same connectionId always re-opened a standalone discovery client and re-ran `belongsToCluster` and `discoverCluster `even though `clients` already held a committed cluster client for that connectionId.

Changes:
- Refactor `connectToValkey` to be more readable and remove duplicate code.
-  Dedupe cluster clients using `clusterId`.
    - `inFlightClusterClients` maps clusterIds to Promise<GlideClusterClient> ensuring different connections connecting to the same cluster (even with different connectionIds) connect to the same ClusterClient.
- Serialize concurrent `connectToValkey` calls for same connectionId.
    - Chain concurrent calls and reuse existing connections.
- Enable parallel execution of integration tests.


apps/server/src/connection.ts
```diff
@@ -203,10 +313,84 @@ export async function belongsToCluster(client: GlideClient): Promise<boolean> {
...
+    .map(([id]) => id)
+
+  try { existingClusterConnection.client.close() } catch (error) {
+    console.error(`Error closing stale client for ${existingClusterConnection.clusterId}:`, error)
+  }
+  // Update map with new client
+  sharedIds.forEach((id) => clients.set(id, { client: newClusterClient!, clusterId: existingClusterConnection.clusterId }))
+}
+
+async function commitClusterConnection(
```
arseny-kostenko, line 382: 9 positional arguments is smelling, make it a context object

apps/server/src/connection.ts
```diff
@@ -256,137 +444,71 @@ export async function discoverCluster(client: GlideClient | GlideClusterClient,
...
-  } catch (err) {
-    console.error("Error connecting to Valkey", err)
-    const errorMessage = err instanceof Error ? err.message : String(err)
-    ws.send(
-      JSON.stringify({
-        type: VALKEY.CONNECTION.connectRejected,
-        payload: {
-          errorMessage,
-          connectionId,
+function sendClusterConnectFulfilled(
```
arseny-kostenko: make it a context object instead of a sequence of arguments

apps/server/src/connection.ts
```diff
@@ -65,72 +107,140 @@ export async function connectToValkey(
...
-    )
-    return standaloneClient
+      subscribe(payload.connectionId, ws)
+
+      await sendStandaloneConnectFulfilled(standaloneClient, connectionId, ws)
+
+      shouldCloseStandaloneClientOnError = false
+      return standaloneClient
 
+    } finally {
```
ravjotbrar: As discussed we likely don't need a finally block to close the standAlone on error, since this should be handled in the catch block

arseny-kostenko (approved):

PR #332, opened 2026-04-30
Fix overshooting key limit and commandlogs limit

ravjotbrar:
## Description

- Command logs were limited at rows, but are returned as values
- In a large cluster all nodes are scanning at once and were overshooting key limit
- Add error handling and documentation if commandlog isn't available

### Change Visualization


apps/metrics/src/init-collectors.js
```diff
@@ -148,7 +148,14 @@ const setupCollectors = async (client, cfg) => {
         },
       }
 
-      const rows = await fn()
+      let rows
```
ArgusLi, line 151: How about we do something like
```typescript
const rows = await (async () => {
    try {
        return await fn()
    } catch (e) {...
```

apps/server/src/keys-browser.ts
```diff
@@ -241,8 +241,7 @@ async function scanCluster(
       const keys = value[1]
 
       keys.forEach((k) => {
-        allKeys.add(k)
-        if (allKeys.size >= limit) return
```
ArgusLi, line 245: What's the purpose for not returning here?

ravjotbrar: The return was kind of misleading. This actually skips adding the keys if the limit is reached

apps/metrics/src/init-collectors.js
ravjotbrar: Yes cleaner, making the change

ravjotbrar: Actually, this would change behaviour since the return only exits the anonymous function not the entire map, so we'll need to keep it

ArgusLi: I was going to suggest
``` typescript
const rows = await fn().catch((err) => {
    ....
    return
})
```

But the nd closing makes it impossible.

ArgusLi (dismissed):

apps/server/src/actions/commandLogs.ts
```diff
@@ -126,20 +126,22 @@ export const commandLogsRequested = withDeps<Deps, void>(
       return
     }
 
-    const aggregatedRows = R.sort(
-      R.descend((row: { values?: Array<{ duration_us?: number; size?: number }>; ts: number }) => {
-        const v = row.values?.[0]
-        return v?.duration_us ?? v?.size ?? row.ts
-      }),
-      results.flatMap((r) => r.rows.map((row) => ({ ...row, nodeId: r.nodeId }))),
+    const allValues = results
```
arseny-kostenko: if you don't need `allValues` anywhere else, I suggest making a pipe (and flattening the flatMaps):
```
const sortedValues = R.pipe(
  R.chain(({ rows, nodeId }) => rows.map((row) => ({ row, nodeId }))),
  R.chain(({ row, nodeId }) => (row.values ?? []).map((v) => ({ ...v, nodeId }))),
  R.sort(
    R.descend(
      (v: { duration_us?: number; size?: number; ts: number }) =>
        v.duration_us ?? v.size ?? v.ts,
    ),
  ),
)(results)
```

ravjotbrar: That's cleaner making the change

arseny-kostenko (approved):

PR #355, opened 2026-05-13
feat: add key analysis page with big key and prefix memory features

jackiesre721:
Add a new Key Analysis page with two diagnostic tools for identifying memory-heavy keys and understanding memory distribution by key prefix.

Server-side changes:
- Add key-analysis.ts: core scanning logic using SCAN command to iterate keys, collecting MEMORY USAGE, TYPE, TTL, and collection size (HLEN, LLEN, SCARD, ZCARD, XLEN) for each key. Uses pLimit(20) for concurrent enrichment with configurable scan limit (default 10,000 keys).
- Add actions/key-analysis.ts: WebSocket action handler using withDeps pattern, registered in server index.ts handlers map.
- Add KEY_ANALYSIS namespace to common/src/constants.ts with four action types: analysisRequested, analysisProgress, analysisFulfilled, analysisError.
- Supports both GlideClient (standalone) and GlideClusterClient (cluster) with per-node cursor tracking for cluster deployments.
- Sends real-time progress updates during scanning and enriching phases.
- Error handling follows existing pattern: send analysisError then check for ConnectionError/TimeoutError/ClosingError.

Frontend state management:
- Add keyAnalysisSlice.ts: Redux slice with connectionId-keyed state, four reducers (requested/progress/fulfilled/error), and eight selectors for status, keys, totals, progress, and error.
- Add keyAnalysisEpic.ts: RxJS epic forwarding analysisRequested actions via WebSocket. Registered in rootEpic.ts and store.ts.

Frontend UI:
- Add KeyAnalysis.tsx: main page component with AppHeader, four StatCards (Total Keys, Scanned, Total Memory, Largest Key), TabGroup toolbar for switching between Big Keys and Prefix Memory tabs, progress bar with percentage during scanning, and refresh button.
- Add BigKeysTable.tsx: sortable table (Key Name, Type, Memory, Size, TTL) with click-to-sort headers, default sorted by memory desc. Includes a ChartModal with Recharts horizontal BarChart showing top 20 keys.
- Add PrefixMemoryTable.tsx: client-side prefix aggregation by colon delimiter with configurable segment depth (1/2/3). Columns: Prefix, Total Memory, Key Count, % of Total, Type Breakdown. Includes donut chart (Recharts PieChart with innerRadius) for memory distribution.
- Add sidebar navigation item (BarChart3 icon) and routes for both standalone (/:id/key-analysis) and cluster (/:clusterId/:id/key-analysis).

Files changed: 13 files, +994 lines (7 new, 6 modified)

## Description

### Change Visualization

### DIFF
--- a/apps/frontend/src/components/key-analysis/BigKeysTable.tsx
+++ b/apps/frontend/src/components/key-analysis/BigKeysTable.tsx
@@ -1,0 +1,198 @@
+import { useState, useMemo } from "react"
+import { BarChart3 } from "lucide-react"
+import {
+  BarChart,
+  Bar,
+  XAxis,
+  YAxis,
+  Tooltip,
+  ResponsiveContainer,
+  CartesianGrid,
+} from "recharts"
+import { formatBytes } from "@common/src/bytes-conversion"
+import { convertTTL } from "@common/src/ttl-conversion"
+import { TableContainer } from "@/components/ui/table-container"
+import {
+  SortableTableHeader,
+  StaticTableHeader,
+  type SortOrder,
+} from "@/components/ui/sortable-table-header"
+import { Badge } from "@/components/ui/badge"
+import { Button } from "@/components/ui/button"
+import { ChartModal } from "@/components/ui/chart-modal"
+import { Typography } from "@/components/ui/typography"
+import type { AnalyzedKeyInfo } from "@/state/valkey-features/key-analysis/keyAnalysisSlice"
+
+interface BigKeysTableProps {
+  keys: AnalyzedKeyInfo[]
+  loading: boolean
+}
+
+type SortField = "name" | "type" | "memoryUsage" | "collectionSize" | "ttl"
+
+export function BigKeysTable({ keys, loading }: BigKeysTableProps) {
+  const [sortField, setSortField] = useState<SortField>("memoryUsage")
+  const [sortOrder, setSortOrder] = useState<SortOrder>("desc")
+  const [chartOpen, setChartOpen] = useState(false)
+
+  const handleSort = (field: SortField) => {
+    if (sortField === field) {
+      setSortOrder(sortOrder === "asc" ? "desc" : "asc")
+    } else {
+      setSortField(field)
+      setSortOrder("desc")
+    }
+  }
+
+  const sortedKeys = useMemo(() => {
+    const multiplier = sortOrder === "asc" ? 1 : -1
+    return [...keys].sort((a, b) => {
+      switch (sortField) {
+        case "name":
+          return multiplier * a.name.localeCompare(b.name)
+        case "type":
+          return multiplier * a.type.localeCompare(b.type)
+        case "memoryUsage":
+          return multiplier * (a.memoryUsage - b.memoryUsage)
+        case "collectionSize":
+          return multiplier * ((a.collectionSize ?? 0) - (b.collectionSize ?? 0))
+        case "ttl":
+          return multiplier * (a.ttl - b.ttl)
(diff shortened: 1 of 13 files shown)


arseny-kostenko:
Thanks for the contribution. I like the intention, it's a useful feature, but I don’t think this should land in its current shape. This seems to conflict with several expectations in `CONTRIBUTING.md` and also raises some architecture/correctness concerns.

### Process / contribution concerns

- You're expected to provide screenshots/video of features adding new UI and/or updating existing UI. Did you familiarize yourself with our contributing guidlines?
- This is a new diagnostic feature with architecture impact, so it likely should have gone through the RFC process before implementation.
- The PR adds new WebSocket actions and a new user-facing UX surface, but I don’t see accompanying docs updates.
- Given the amount of duplicated logic, generic scaffolding, and unclear design tradeoffs, this also raises concerns under the generative-AI contribution guidance. Could you explain which existing key-browser and metrics-server paths you evaluated, whether any AI tooling was used, and why this separate implementation is the right approach?

### Architecture concerns

- This puts keyspace-wide scan/aggregation work in `apps/server`, but this seems closer to the metrics/collector boundary than the main admin backend.
- The implementation adds a second keyspace crawler instead of reusing/refactoring the existing key browser primitives in `keys-browser.ts`.
- We now have duplicated standalone scan, cluster scan, `TYPE`, `TTL`, `MEMORY USAGE`, collection-size lookup, cluster primary routing, scan cursor handling, and progress semantics. These implementations will drift over time.

### React / frontend concerns

- `KeyAnalysis.tsx` dispatches `analysisRequested` from `useEffect`, but `CONTRIBUTING.md` explicitly says not to use `useEffect` for application data flow.
- Loading the route automatically starts a potentially expensive keyspace scan. I’d expect this to be explicitly user-triggered, gated, or handled by the metrics layer.
- The UI presents sampled/capped analysis as if it were authoritative database-wide analysis. This capping is also quite arbitrary and not configurable.
- “Total Memory” is actually memory summed from scanned/analyzed keys, not total database memory.
- Prefix percentages are based on the analyzed sample, not the full keyspace.
- Prefix grouping is an arbitrary frontend heuristic rather than a backend/collector aggregation.

### Backend / correctness concerns

- The default scan path can issue a large number of commands from the admin backend: `SCAN`, then per-key `TYPE`, `MEMORY USAGE`, `TTL`, and sometimes `HLEN` / `LLEN` / `SCARD` / `ZCARD` / `XLEN`.
- The default limits and sampling behaviour are not clearly surfaced to the user, so the results can be misleading.
- There is no cancellation or stale-result handling if the user navigates away, switches connections, disconnects, or starts another analysis.
- Error handling collapses failures into `type: "unknown"`, `memoryUsage: 0`, and `ttl: -1`, which can silently undercount memory and hide command/permission/cluster-routing issues.
- Type handling is not consistent with the existing key browser path, including existing handling for string/JSON-ish key types and size semantics.
- Cluster-wide analysis semantics need more thought. Progress, total key estimates, and aggregation across primaries should be handled consistently and probably shared with or owned by the metrics path.
- The new frontend files are flagged by GitHub as containing hidden/bidirectional Unicode text, which should be removed or explained before merge.

I'd suggest pausing this implementation and starting with an RFC/design proposal. From there, I think this should be reworked around shared key-browser primitives and/or moved into the metrics-server/collector path, with explicit user-triggering, cancellation/stale-result handling, docs updates, and much clearer sample-vs-total semantics in the UI.

ravjotbrar:
Closing this PR as there has been no activity in the past month

PR #358, opened 2026-05-25
restricting destructive commands that cause server issues

### LINKED ISSUE #284: Restrict desructive commands
The Send Command interface currently allows execution of destructive commands such as FLUSHALL, FLUSHDB, DEBUG, SHUTDOWN, and others that can cause irreversible data loss or disrupt the server.

In addition, commands like KEYS * should also be restricted.

Expected behavior:
Destructive commands should either be blocked outright or require an explicit confirmation step 
before execution.

Suggested commands to restrict:
- FLUSHALL / FLUSHDB — deletes all keys
- SHUTDOWN — stops the server
- DEBUG — can cause crashes or data corruption
- CONFIG RESETSTAT / CONFIG REWRITE — modifies server configuration
- SLAVEOF / REPLICAOF — changes replication topology
- CLUSTER RESET — resets cluster state

Proposed solution:
Maintain a blocklist of dangerous commands on the server side. For commands that have legitimate use cases (e.g. FLUSHDB in a dev environment), show a confirmation dialog before sending.

nassery318:
## Description

Restricting destructive commands with two methods 

1. defining BLOCKED_COMMANDS (hard-blocked) that causes server crashes (DEBUG and SHUTDOWN)
2. defining CONFIRM_COMMANDS (require user confirmation) can still cause crashes and data loss

### Change Visualization

### DIFF
--- a/apps/frontend/src/components/send-command/CommandConfirmDialog.tsx
+++ b/apps/frontend/src/components/send-command/CommandConfirmDialog.tsx
@@ -1,6 +1,8 @@
 import * as Dialog from "@radix-ui/react-dialog"
 import { AlertTriangle, X } from "lucide-react"
+import { useState } from "react"
 import { Button } from "@/components/ui/button"
+import { Input } from "@/components/ui/input"
 import { Typography } from "@/components/ui/typography"
 import { Alert, AlertDescription } from "@/components/ui/alert"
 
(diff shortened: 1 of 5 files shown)


nassery318:
https://github.com/user-attachments/assets/fe317247-374b-4e41-b2c1-af54e5b0fb7b

arseny-kostenko (approved):

PR #364, opened 2026-06-01
replaced blocking key info commands with pagination for different key…

### LINKED ISSUE #244: ENHANCEMENT: Replace blocking key info commands
# Problem
 
In `apps/server/src/keys-browser.ts` the element commands used are blocking commands. While [Hash and Set key types have been updated to use Scan style commands instead](https://github.com/valkey-io/valkey-admin/pull/232), upgrading the remaining commands would lead to a performance improvement on the monitored nodes.

# Solution

Upgrade the following element commands

Key Type | Old Command | New Command | Type
-- | -- | -- | --
List | LRANGE |   | Pagination
Zset | ZRANGE | ZRANGE ... BYSCORE ... BYLEX | Pagination with same score de-duplication
Stream | XRANGE |   | Pagination
rejson-rl | JSON.GET |   | Split based on if JSON is object or array

nassery318:
… types

## Description

Replaced blocking key info commands for (List, Stream, Zset and JSON) with pagination 

Issue: ENHANCEMENT: Replace blocking key info commands #244

### Change Visualization

### DIFF
--- a/apps/server/src/keys-browser.ts
+++ b/apps/server/src/keys-browser.ts
@@ -180,10 +180,10 @@ async function getPaginatedZSetInfo(
     // members lexicographically with BYLEX, resuming just past the last one we kept.
     // BYLEX can't return scores (WITHSCORES is rejected), but every member in this
     // range shares `score`, so we re-attach it ourselves.
-    const drainTiedScore = async (score: string, afterMember: string, keptAtScore: number): Promise<void> => {
+    const drainTiedScore = async (score: string, lastScannedAtScoreKey: string, keptAtScore: number): Promise<void> => {
       const tiedCount = await client.customCommand(["ZCOUNT", keyInfo.name, score, score]) as number
       let remaining = tiedCount - keptAtScore
-      let lexCursor = `(${afterMember}` // exclusive, so we don't re-read the last kept member
+      let lexCursor = `(${lastScannedAtScoreKey}` // exclusive, so we don't re-read the last kept member
       while (remaining > 0) {
         const members = await client.customCommand([
           ...commands.elementsCmd, lexCursor, "+", "BYLEX", "LIMIT", "0", VALKEY_CLIENT.ELEMENT_PAGE_SIZE.toString(),
@@ -213,12 +213,12 @@ async function getPaginatedZSetInfo(
       if (addMembers(page) === 0) {
         // A full page with no new members means this score is tied beyond the page size.
         // Drain the rest of it lexicographically with BYLEX, then resume past the score.
-        const keptAtScore = results.reduce((n, m) => (String(m.value) === lastScore ? n + 1 : n), 0)
-        await drainTiedScore(lastScore, page[page.length - 1].key, keptAtScore)
-        const next = await client.customCommand(buildZRangeByScoreArgs(`(${lastScore}`)) as ZSetMember[]
-        if (!next || next.length === 0) break
-        addMembers(next)
-        lastPage = next
+        const scannedAtScoreCount = results.reduce((n, m) => (String(m.value) === lastScore ? n + 1 : n), 0)
+        await drainTiedScore(lastScore, page[page.length - 1].key, scannedAtScoreCount)
+        const matchingLastScoreCount = await client.customCommand(buildZRangeByScoreArgs(`(${lastScore}`)) as ZSetMember[]
+        if (!matchingLastScoreCount || matchingLastScoreCount.length === 0) break
+        addMembers(matchingLastScoreCount)
+        lastPage = matchingLastScoreCount
       } else {
         lastPage = page
       }
--- a/apps/server/src/keys-browser.ts
+++ b/apps/server/src/keys-browser.ts
@@ -74,6 +74,192 @@ async function getScanKeyInfo(
   }
 }
 
+async function getPaginatedListInfo(
+  client: GlideClient | GlideClusterClient,
+  keyInfo: EnrichedKeyInfo,
+  commands: { sizeCmd: string },
+): Promise<EnrichedKeyInfo> {
+  try {
+    const [firstPage, collectionSize] = await Promise.all([
+      client.customCommand(["LRANGE", keyInfo.name, "0", (VALKEY_CLIENT.ELEMENT_PAGE_SIZE - 1).toString()]),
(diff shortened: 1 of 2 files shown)


apps/server/src/keys-browser.ts
```diff
@@ -132,16 +318,15 @@ export async function getKeyInfo(
     // Get collection size and elements for each type
     const elementCommands: Record<
       string,
-      { sizeCmd: string; elementsCmd: string[] }
+      { sizeCmd: string; elementsCmd?: string[] }
```
ArgusLi: Instead of making this optional and setting it in the paginated commands, can we just set the elementsCmd base command here? i.e. for list, have elementsCmd be just ["LRANGE", key]

Then in the actual function, add to the elementsCmd.

ArgusLi: The reason being, this allows for a single source for what commands are being used, and also prevents elementsCmd being optional, which leads to an additional check.

apps/server/src/keys-browser.ts
```diff
@@ -74,6 +74,192 @@ async function getScanKeyInfo(
...
+    const [firstPage, collectionSize] = await Promise.all([
+      client.customCommand(["XRANGE", keyInfo.name, "-", "+", "COUNT", VALKEY_CLIENT.ELEMENT_PAGE_SIZE.toString()]),
+      client.customCommand([commands.sizeCmd, keyInfo.name]),
+    ]) as [GlideReturnType[][], number]
+
+    const results: GlideReturnType[][] = [...firstPage]
+
+    if (firstPage.length === VALKEY_CLIENT.ELEMENT_PAGE_SIZE) {
+      const lastId = firstPage[firstPage.length - 1][0] as string
+      let cursor = `(${lastId}`
```
ArgusLi, line 124: Can we add a comment here stating the ( is intentional, and link to the doc page?

apps/server/src/keys-browser.ts
```diff
@@ -74,6 +74,192 @@ async function getScanKeyInfo(
...
+
+type ZSetMember = { key: string; value: string | number }
+
+async function getPaginatedZSetInfo(
+  client: GlideClient | GlideClusterClient,
+  keyInfo: EnrichedKeyInfo,
+  commands: { sizeCmd: string },
+): Promise<EnrichedKeyInfo> {
+  // Builds one "page" query: members from `min` score upward, capped at PAGE_SIZE
+  const range = (min: string): string[] => [
```
ArgusLi: Can a more descriptive variable name be used instead of range here?

apps/server/src/keys-browser.ts
```diff
@@ -74,6 +74,192 @@ async function getScanKeyInfo(
...
+          added++
+        }
+      }
+      return added
+    }
+
+    addMembers(firstPage)
+    let lastPage = firstPage
+
+    // Keep paging until we have every member, or a page comes back not full (the end).
```
ArgusLi, line 205: I can see a potential bug if the number of ZSet members with the same score exceeds the Page Size. Since we're using score to advance the page, we could get stuck. 

e.g. page size is 3 and the ZSet looks like [(0, "a"), (0, "b"), (0, "c"), (0, "d")]
In this case, you'll be returning the same page of [(0, "a"), (0, "b"), (0, "c")].

Let's use BYLEX to break the tie in that case.

apps/server/src/keys-browser.ts
```diff
@@ -74,6 +74,192 @@ async function getScanKeyInfo(
...
+    return { ...keyInfo, collectionSize, elements: results }
+  } catch (err) {
+    console.log(`Could not get elements for key ${keyInfo.name}:`, err)
+    return { ...keyInfo, elementsWarning: VALKEY_CLIENT.MESSAGES.NOT_READABLE }
+  }
+}
+
+// JSON queries give back the value inside an array, like [value]. This returns just the value
+function unwrapJsonPathResult<T>(result: unknown): T {
+  return (Array.isArray(result) ? result[0] : result) as T
```
ArgusLi, line 236: Can this function name be more descriptive? i.e. getJsonPathType

ArgusLi (changes requested):

apps/server/src/keys-browser.ts
nassery318: addressed

apps/server/src/keys-browser.ts
```diff
@@ -181,12 +207,15 @@ async function getPaginatedZSetInfo(
...
-      const page = await client.customCommand(range(lastScore)) as ZSetMember[]
+      const page = await client.customCommand(buildZRangeByScoreArgs(lastScore)) as ZSetMember[]
       if (!page || page.length === 0) break
 
       if (addMembers(page) === 0) {
-        // Page had no new members, so skip past this score
-        const next = await client.customCommand(range(`(${lastScore}`)) as ZSetMember[]
+        // A full page with no new members means this score is tied beyond the page size.
+        // Drain the rest of it lexicographically with BYLEX, then resume past the score.
+        const keptAtScore = results.reduce((n, m) => (String(m.value) === lastScore ? n + 1 : n), 0)
```
ArgusLi: Can the variable name be more descriptive here? i.e. matchingLastScoreCount

apps/server/src/keys-browser.ts
```diff
@@ -173,6 +175,30 @@ async function getPaginatedZSetInfo(
       return added
     }
 
+    // When a single score has more tied members than fit in one page, BYSCORE can't
+    // advance without dropping the rest of the tie. Finish that score by walking its
+    // members lexicographically with BYLEX, resuming just past the last one we kept.
+    // BYLEX can't return scores (WITHSCORES is rejected), but every member in this
+    // range shares `score`, so we re-attach it ourselves.
+    const drainTiedScore = async (score: string, afterMember: string, keptAtScore: number): Promise<void> => {
```
ArgusLi: Can we rename keptAtScore to be more descriptive here? i.e. scannedAtScoreCount?

ArgusLi: On the same note, can afterMember be more descriptive? i.e. lastScannedAtScoreKey?

ArgusLi (changes requested):
Some small changes for readability. Overall I like the approach to break the tie. There's probably something more performant, but this works.

ArgusLi (approved):

PR #367, opened 2026-06-08
Remove db on all boundaries between connectionId and metricsId

ArgusLi:
## Description

This PR fixes the boundaries between client connectionIds and metrics server Ids by stripping db at all instances client connectionIds are used to access metrics instances.

Client connectionIds keys carry `db` because each (host, port, db) is a distinct user-visible connection with its own Glide client.

Metrics server keys do not carry `db`. A metrics process is one OS process per Valkey node (host, port); the data it collects is node level, not db-scoped.

### DIFF
--- a/apps/server/src/__tests__/connection.test.ts
+++ b/apps/server/src/__tests__/connection.test.ts
@@ -13,7 +13,7 @@ import {
 import { resolveHostnameOrIpAddress, dns } from "../utils"
 import { checkJsonModuleAvailability } from "../check-json-module"
 import { ConnectionDetails } from "../actions/connection"
-import { type MetricsServerMap } from "../metrics-orchestrator"
+import { toMetricsNodeId, type MetricsServerMap } from "../metrics-orchestrator"
 import { _reset as resetNodeWatchers } from "../node-watchers"
 
 const DEFAULT_PAYLOAD = {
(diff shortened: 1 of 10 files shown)


ravjotbrar:
I personally don't think db should be a part of the connectionID. There's commands provided by Valkey to switch DBs within the same connection and that renders our connectionId invalid or at the very least, misleading. 

https://valkey.io/commands/select/

ArgusLi:
> I personally don't think db should be a part of the connectionID. There's commands provided by Valkey to switch DBs within the same connection and that renders our connectionId invalid or at the very least, misleading.
> 
> https://valkey.io/commands/select/

The reason why I decided to implement it this way is because Valkey Glide client configs [set a databaseId](https://glide.valkey.io/languages/nodejs/api/interfaces/BaseClient.BaseClientConfiguration.html#databaseid) (for both standalone and cluster). If left unset, it defaults to 0. Yes you could modify the state to use a different database id by using SELECT, but on reconnection, the client will use what was in the config to begin with.

This creates a second issue. With our current architecture that shares clients for different Valkey Admin instances accessing the same node, if one instance of Admin modifies the client's state to use a different database (using SELECT), the other clients using the same connection will be directly affected.

The solution that I came up with is to have db be part of the connectionId. In a future PR that I have planned, I will disable a bunch of commands (including SELECT) that can modify the state in a way that prevents the UI from knowing. Part of that PR would be to create a way to change the Database in the UI (apart from editing it in the connection tab). This will create a new client that is keyed with the database id.

Happy to discuss.

nassery318:
In case you want to block commands (i.e. SELECT or similar) you can utilize the command-restrictions.ts or something along the same lines.

ravjotbrar (approved):

PR #369, opened 2026-06-10
Refactor info from server to metrics

### LINKED ISSUE #210: Refactor dashboard data to metrics server
Currently, we are using our backend server to do the job of our metrics server by using it to get dashboard metrics. We should get these metrics by calling the appropriate endpoint in the metrics server.

The code that we need to change is contained in set-dashboard-data.ts. We also have to create an endpoint in the metrics server to get info.

nassery318:
## Description

1. Moved info metrics from Server to metrics
2. fix the metrics server uri registration issue
3. display skeleton while metrics uri is registering
4. display alert if the metrics server did not start

### Change Visualization

https://github.com/user-attachments/assets/2b1062cb-10e9-459d-a613-c00a8c7f4596


arseny-kostenko (approved):

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -522,6 +525,58 @@ export const monitorEpic = () =>
     ignoreElements(),
   )
 
+// metric server not ready retry epic for dashboad data
```
ArgusLi, line 528: nit: dashboard

apps/frontend/src/state/valkey-features/info/infoSlice.ts
```diff
@@ -67,6 +69,7 @@ interface InfoSliceState {
 
 const createInitialConnectionState = (): ConnectionState => ({
   error: null,
+  metricsStarting: false,
```
ArgusLi, line 72: Since it's true while the metrics server is still starting up, should this be true?

apps/frontend/src/state/epics/valkeyEpics.ts
nassery318: good point of clarification , will adjust in next PR

apps/frontend/src/state/valkey-features/info/infoSlice.ts
nassery318: metricsStarting starts false on purpose cause the reducers always set it themselves, so the initial value is never read, changing it to true would do nothing. below is the flow:

  1. When a connection first starts, the reducer creates its state with metricsStarting: false.
  2. But in that same step, the reducer immediately sets the real value:
    - setData → sets it to false (data arrived)
    - error (not ready) → sets it to true

nassery318: but as you mentioned, true reads better as things are getting ready to start. I will adjust it to true.

PR #373, opened 2026-06-19
 Improve error handling for unsupported server commands

ravjotbrar:
### Summary
  
  When connecting to a server that doesn't support certain commands (e.g., ElastiCache
  Serverless), Valkey Admin previously showed confusing errors, blank UI states, or
  falsely reported features as active. This PR surfaces clear error messages to the user
  and adds a connection-time warning for serverless caches.
  
###  Problem
  
  - Monitor showed "Active" even when the MONITOR command was rejected by the server
  - Command Logs crashed with "Cannot read properties of null" when COMMANDLOG was
  unsupported
  - No indication to the user that they were connecting to a limited environment
  
###  Changes
  
  - Monitor: The metrics server now waits for the first MONITOR cycle to succeed before
  reporting "running." If the command is rejected, the error propagates to the frontend
  and displays in the Activity view.
  - Command Logs: The metrics server now stores startup failures and returns a proper
  error response instead of crashing when the collector never initialized. Fixed a key
  mismatch between the request type and collector name that caused the error check to be
  bypassed.
  - Serverless detection: On connection, if the hostname contains .serverless., a toast
  warns the user which features are unavailable.
  - Memory Usage: When MEMORY USAGE is unavailable (e.g., blocked on serverless), key sizes now display "—" instead of "0 B". The total memory calculation and distribution chart skip keys with unknown sizes rather than counting them as zero.
  - General: Unknown command errors in the monitor stream now terminate the retry loop
  instead of retrying indefinitely.


apps/frontend/src/components/ui/donut-chart.tsx
```diff
@@ -50,7 +50,7 @@ export default function DonutChart() {
   // count and memory usage per each key type
   const keyTypeInfo = keys.reduce((stats, key) => {
     (stats[key.type] ??= { count: 0, totalSize: 0 }).count += 1
-    stats[key.type].totalSize += key.size
+    if (key.size > 0) stats[key.type].totalSize += key.size
```
ArgusLi, line 53: Is this guard to prevent if key.size < 0 or is this an optimization?

ArgusLi (dismissed):
Just a small question. I like the way init-collectors works way more now.

apps/frontend/src/components/ui/donut-chart.tsx
ravjotbrar: If MEMORY commands aren't supported, we set the size of a key to -1, since it has to be a number. In the frontend, we detect this and display the size as —

apps/metrics/src/effects/monitor-stream.js
```diff
@@ -3,66 +3,87 @@ import { exhaustMap, catchError, map } from "rxjs"
...
+/**
+ * Phase 1: Connect and issue MONITOR command.
+ * Returns the client and monitor handle for use in Phase 2.
+ * Throws immediately if the command is unsupported.
+ */
+export const connectMonitor = async () => {
+  const { host, port, username, tls } = getConnectionOptions()
+  const password = await getPassword()
+  const client = new Valkey({ host, port, username, password, tls })
+  client.on("error", (err) => console.error("[monitor] ioredis client error:", err.message))
```
ArgusLi: nit: should this be iovalkey?

apps/metrics/src/init-collectors.js
```diff
@@ -59,62 +64,52 @@ const startMonitor = (cfg) => {
...
+    complete: () => {
+      updateCollectorMeta(monitorEpic.name, {
+        completedAt: Date.now(),
+        isRunning: false,
+      })
+      console.debug(`[${monitorEpic.name}] monitor completed`)
+    },
   })
+
+  monitorStopper = async () => {
```
ArgusLi, line 103: Does this split from a singular updateCollectorMeta to 2 updateCollectorMetas have any upstream impact? Is there a reason why this needed to be split?

apps/metrics/src/effects/monitor-stream.js
ravjotbrar: The library is unchanged from the initial code, which is iovalkey. Before it used to be called monitorClient and I can revert that piece?

ArgusLi: I'm more addressing the error message which says ioreddis. The library is now iovalkey so should we update the error message?

apps/metrics/src/init-collectors.js
ravjotbrar: Good question. I split runMonitorOnce into two separate functions because I needed a way to return immediately if the monitor command wasn't supported. Before the code would need to await the entire duration of monitor for runMonitorOnce to return. Now we can await connectMonitor for an immediate result.

apps/metrics/src/effects/monitor-stream.js
ravjotbrar: Missed that, thanks for catching

ArgusLi (approved):

PR #375, opened 2026-06-22
Refactor connectionId, nodeId, clusterId

ArgusLi:
## Description

Previously the metrics state was keyed by the db-suffixed `connectionId `(<host>-<port>-db<N>) even though the metrics process is unique per Valkey node. That mismatch caused state to be stored and looked up under inconsistent keys.

This branch establishes a single id space for metrics and re-keys all node-level state to the db-less `nodeId` (<host>-<port>).

## Bug fixes
- Fixed double hot-key response.
- Fixed standalone hot-keys showing no result.

## Tooling
- Renamed `eslint.config.js` to `eslint.config.mjs` so the ESM-only config loads (the plugins are ESM-only and the root package is CommonJS).
- Make `npm run lint` automatically fix.


apps/server/src/actions/config.ts
```diff
@@ -10,56 +9,59 @@ interface ParsedResponse  {
...
-        const normalizedError = {
-          success: false,
-          message: "Metrics server URI not found",
-          data: {},
-        }
-        sendUpdateError(ws, connectionId, normalizedError)
-        return
+    if (typeof clusterId === "string") {
+      const nodeIds = Object.keys(clusterNodesRegistry[clusterId] ?? {}).filter((id) => metricsServerMap.has(id))
+      const responses = await Promise.all(
```
arseny-kostenko, line 45: a node may be in a semi-alive state like when it's a failover or something, and it will fail the promise, so maybe worth switching to `allSettled` instead?

arseny-kostenko: in other words, if one node is to be killed anyway — do we need to fail the operation that succeeded in all alive nodes?

apps/server/src/actions/config.ts
```diff
@@ -10,56 +9,59 @@ interface ParsedResponse  {
...
+        nodeIds.map((nodeId) => postConfigToNode(metricsServerMap.get(nodeId)?.metricsURI, config)),
+      )
+      const firstFailure = responses.find((r) => !r.success)
```
arseny-kostenko, line 48: subsequently, just getting the entire list of failing nodes instead of re-sending an update just to discover another failing node?

and then use filter instead of findFirst?

apps/server/src/actions/config.ts
```diff
@@ -10,56 +9,59 @@ interface ParsedResponse  {
...
+      if (firstFailure) {
+        sendUpdateError(ws, { clusterId }, firstFailure)
+      } else {
+        // All nodes responses are the same so we use the first.
+        sendUpdateFulfilled(ws, { clusterId }, responses[0] ?? { success: true, message: "", data: {} })
```
arseny-kostenko, line 53: updating a config is a rarely used operation so we don't have to optimize for network traffic. I'd prefer to have an explicit node: status response

apps/server/src/actions/hotkeys.ts
```diff
@@ -55,26 +54,19 @@ const sendHotKeysError = (
 export const hotKeysRequested = withDeps<Deps, void>(
   async ({ ws, metricsServerMap, action, clusterNodesRegistry }) => {
     const { connectionId, clusterId, lfuEnabled, clusterSlotStatsEnabled } = action.payload
-    
+
     const nodes =
-      typeof clusterId === "string"
-        ? clusterNodesRegistry[clusterId]
+      typeof clusterId === "string"
```
arseny-kostenko, line 59: did lint add trailing spaces?

apps/server/src/actions/config.ts
ArgusLi: I think that's a good change. For this PR, my focus was on retaining existing logic and just ensuring my changes don't break anything. Considering #373 is waiting for this PR so that the bug where errors are not propagated to the right ID is fixed, I would prefer to not add this and the additional handling in this PR, and add it in its own PR right after.

apps/server/src/actions/hotkeys.ts
ArgusLi: It appears to have.

apps/frontend/src/components/activity-view/ActivityView.tsx
```diff
@@ -48,18 +49,23 @@ export const ActivityView = () => {
...
-  const hotKeysData = useSelector((state: RootState) => selectHotKeys(hotKeysId)(state))
-  const hotKeysStatus = useSelector((state: RootState) => selectHotKeysStatus(hotKeysId)(state))
-  const hotKeysErrorMessage = useSelector((state: RootState) => selectHotKeysError(hotKeysId)(state))
-  const hotKeysNodeErrors = useSelector((state: RootState) => selectHotKeysNodeErrors(hotKeysId)(state))
-  const hotKeysLastCollectedAt = useSelector((state: RootState) => selectHotKeysLastCollectedAt(hotKeysId)(state))
-  const monitorRunning = useSelector(selectMonitorRunning(id!))
+  // `targetId` keys node-level metrics state: `clusterId` for a cluster, else
+  // the db-less `nodeId`. (Connection-scoped state below still uses `id`, the
+  // db-suffixed connectionId.)
+  const nodeId = toNodeId(id!)
```
ravjotbrar, line 55: Does it make sense to just store nodeId in frontend state instead of stripping the db suffixed ID every time we need it?

apps/frontend/src/components/ui/monitor-warning-banner.tsx
```diff
@@ -32,28 +30,27 @@ export function MonitorWarningBanner() {
...
-    const grouped = new Set(Object.values(groups).flat().map((c) => c.connectionId))
+    // Monitor entries are grouped by clusterId on the cluster path.
+    // Standalone entries have no clusterId.
     for (const conn of runningConnections) {
-      if (!grouped.has(conn.connectionId)) standalone.push(conn)
+      if (conn.clusterId) {
+        (groups[conn.clusterId] ??= []).push(conn)
+      } else {
+        standalone.push(conn)
+      }
```
ravjotbrar, line 43: This is much better

apps/frontend/src/components/activity-view/ActivityView.tsx
ArgusLi: We actually are using targetId as the key in the frontend state for node level metrics state. This can be seen in the following functions in the code i.e. `selectCommandLogs()()`, `selectHotKeys()()`...

The reason for keeping the connectionId key state is that we still have things that need to be scoped by db such as the connection details, keys etc.

apps/frontend/src/state/valkey-features/config/configSlice.ts
```diff
@@ -19,19 +21,16 @@ interface MonitorConfig {
   cutoffFrequency: number,
 }
 interface ConfigState {
-  [connectionId: string]: {
-    darkMode: boolean,
-    // Valkey related. TODO: find best way to expose to user
-    keyEvictionPolicy?: KeyEvictionPolicy
-    clusterSlotStatsEnabled?: boolean,
```
ravjotbrar, line 26: I like the cleanup here

apps/server/src/actions/commandLogs.ts
```diff
@@ -42,11 +41,11 @@ type CommandLogsLargeResponse = {
 
 type CommandLogResponse = (CommandLogsLargeResponse | CommandLogsSlowResponse) & { nodeId?: string }
 
-type NodeError = { connectionId: string; error: string }
+type NodeError = { nodeId: string; error: string }
 
 const sendCommandLogsFulfilled = (
   ws: WebSocket,
-  connectionId: string,
+  replyId: AggregateReplyId, // Commandlogs aggregates all cluster nodes into one result
```
ravjotbrar, line 48: Why do we name it as replyId? This would be simpler to comprehend as clusterId if that is what we store commandlogs under.

ArgusLi: This is because sendCommandLogsFulfilled is called by both cluster or standalone. So the replyId can either be clusterId or nodeId.

apps/frontend/src/components/activity-view/ActivityView.tsx
ravjotbrar: I'm more so referring to storing nodeId in connection state, so components can read it directly instead of calling toNodeId every time. But this isn't a blocker if you don't think its worth the time.

ravjotbrar (approved):

apps/frontend/src/components/activity-view/ActivityView.tsx
ArgusLi: Adding nodeId in connection state means that if the connectionId gets changed, we have to ensure nodeId stays consistent, which is an additional thing to manage. toNodeId() is also cheap operation so there's no real performance benefit from caching it directly in connection state. 

So in my mind it's not worth gaining that really small performance gain in exchange for an additional state we have to manage.

PR #376, opened 2026-06-24
End to end implementation of big keys from metrics, to server and to …

nassery318:
…frontend

## Description

Complete implementation  of Big Keys flow from Metrics layer to Server and all the way to Fronted.

https://github.com/user-attachments/assets/2f485d07-045c-4dd3-8d4c-7a97bfad3a81

### Change Visualization

https://github.com/user-attachments/assets/2f485d07-045c-4dd3-8d4c-7a97bfad3a81


apps/server/src/actions/bigkeys.ts
```diff
@@ -0,0 +1,146 @@
+import { type WebSocket } from "ws"
```
ravjotbrar, line 1: Once Argus' merges his refactoring PR, you'll have to merge with main. Look into how he reorganized similar actions and I would recommend implementing the same here.

ravjotbrar (commented):
Overall looks good. I would recommend making big keys configurable similar to other activity results.

apps/frontend/src/components/activity-view/bigkeys/big-keys.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function BigKeys({
+  data, errorMessage, status, nodeErrors, isCluster,
+}: BigKeysProps) {
+  const [sortOrder, setSortOrder] = useState<SortOrder>("desc")
+  const [searchQuery, setSearchQuery] = useState("")
+  const [selectedNode, setSelectedNode] = useState("all")
+
+  if (status === "Pending") return <LoadingState message="Scanning for big keys..." />
+
+  const sorted = R.sort<BigKey>(
```
arseny-kostenko: I'd probably swap sorted and filtered, so you sort what's filtered first. For 50 keys it's not a big deal but I imagine we'll be asked to make 50 configurable (to a much larger number)

arseny-kostenko (dismissed):

arseny-kostenko (approved):

apps/frontend/src/components/activity-view/bigkeys/big-keys.tsx
nassery318: Good point! now we filter first and only sort the matching results, so we avoid sorting the whole dataset.

PR #378, opened 2026-06-26
added big keys configuration for top scan and scan limit

nassery318:
## Description

- Big keys config enabled with configurable scan limit and topN big keys

### Change Visualization

https://github.com/user-attachments/assets/e0b9ca1e-4a98-4912-8910-35b1ae1c5ba4


ravjotbrar (commented):
Overall LGTM. What happens if the user sets topN to larger than the number of keys returned? Do we need to add an upper limit?

nassery318:
> Overall LGTM. What happens if the user sets topN to larger than the number of keys returned? Do we need to add an upper limit?

There is no need for the upper limit. If the limit is larger than the number of results, we just get all of them - R.take/slice return the full list when the limit exceeds its length, and the metrics server treats the limit as a max - so no errors or empty rows.

ravjotbrar (approved):

PR #379, opened 2026-06-29
Add preconfigured metrics server for standalone

### LINKED ISSUE #346: [Future Request] Preconfigured Standalone Server Monitor
Will be cool to have monitor also for Preconfigured Standalone Valkey server that not cluster

ravjotbrar:
## Description

Extracts preconfigured connection logic into explicit methods in the orchestrator. A new startPreconfiguredMetricsServers() entry point checks whether the preconfigured connection is a cluster or standalone, then takes the appropriate path: 
1. for clusters in web mode it discovers topology, populates the registry, and starts the reconcile loop
2. for standalone nodes it starts a single metrics server directly regardless of web or electron modes
3. for Kubernetes mode it's a no-op since sidecars handle metrics collection.
  
reconcileClusterMetricsServers now only diffs the registry against running metrics servers and starts/stops as needed. Discovery is now a setup concern handled before the loop starts. The reconcile loop and NodeInfo type (renamed from ClusterNodeInfo) were moved to the orchestrator to reduce noise in index.ts.

### Change Visualization

### DIFF
--- a/apps/server/src/__tests__/metrics-orchestrator.test.ts
+++ b/apps/server/src/__tests__/metrics-orchestrator.test.ts
@@ -238,11 +238,9 @@ describe("metrics-orchestrator", () => {
   })
 
   describe("reconcileClusterMetricsServers", () => {
-    let connectionDetails: ConnectionDetails
 
     beforeEach(() => {
       metricsServerMap.clear()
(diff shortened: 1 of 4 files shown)


arseny-kostenko (approved):

ArgusLi (approved):
Beautiful

PR #385, opened 2026-07-08
decline helmet default values for "upgrade-insecure-requests"

### LINKED ISSUE #384: Access via hostname other than "localhost", "127.0.0.1" causes TLS problems, starting with "edge"
**Issue:**

Change in behaviour where https is now assumed for non-"localhost" addresses.

**To duplicate:**

podman run -p 8080:8080 --name valkey --rm docker.io/valkey/valkey-admin:1.0.2

I can now access either http://localhost:8080 or via IP address at http://192.168.xxx.yyy:8080

With edge:

podman run -p 8080:8080 --name valkey --rm docker.io/valkey/valkey-admin:edge

I can access http://localhost:8080 but if I try to use the IP or a local hostname it fails with SSL errors as the resources flip to attempt to load via https (not http).

**Origin**

This looks related to the CSR or COOP headers, maybe both.  If it's deemed better to not fix by permitting an environment variable override of trusted local hostname, I think it would be wise to document the limitation.

dbaker-arch:
## Description

Fixes issue 384 - by declining the helmet default Content-Security-Policy values we take only what we declare, and do not pull in the default "upgrade-insecure-requests".

This fixes a regression in behaviour from 1.0.x to edge, not seen when testing is only done against http://localhost:8080 or http://127.0.0.1:8080/

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -81,6 +81,7 @@ app.use((req, res, next) => {
 // Content Security Policy via helmet
 app.use(helmet({
   contentSecurityPolicy: {
+    useDefaults: false,
     directives: {
       defaultSrc: ["'self'"],
       scriptSrc: ["'self'"],


ravjotbrar (commented):
Thanks for digging into this and finding a solution @dbaker-arch! 

One suggestion, if we're disabling defaults, can we explicitly add defaults for frameAncestors to 'self' and scriptSrcAttr to 'none'?

dbaker-arch:
Here's the CSP I extracted from headers
- default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';connect-src 'self';img-src 'self' data:;font-src 'self';object-src 'none';base-uri 'self';form-action 'self'

And a resource I found to help interpret it
- https://csp-evaluator.withgoogle.com/

It flagged two things for attention.
- (script-src) 'self' can be problematic if you host JSONP, AngularJS or user uploaded files.
- (require-trusted-types-for [missing]) Consider requiring Trusted Types for scripts to lock down DOM XSS injection sinks. You can do this by adding "require-trusted-types-for 'script'" to your policy


Indeed the first thing flagged is what you said.  This is, however, problematic to change since we actually feed the application javascript from "self", as seen here in "index.html"
```
<script type="module" crossorigin src="./assets/index-BGQAgxZI.js"></script>
```

It's easy to test this by flipping "self" to "none" and the application breaks, with a "blocked:csp" notice showing for its own javascript.

What's being flagged for awareness is that if we host javascript, we might also host malicious javascript.  We do however host javascript and need to host it, and need to load it, for the app to work at all!



Your other suggestion on frameAncestors isn't so clear.  This dictates whether another page can embed this site within itself.  It feels plausible that we would want to deny it, but it's also unclear on how an end user wants to deploy the app.


I tested and settinging script-src to "none" will break the app.  I've not tested frame-ancestors but feel it might cause silent breakage for users with specific deployment expectations.  This could be an "opt-in" option rather than an application default.


---

I deliberately kept this PR as small as possible for simplicity.  If we were doing any change along with it, I would call for the CSP header in index.html to be removed to avoid confusion - there's no need to set the header in multiple places.

ravjotbrar (approved):

PR #391, opened 2026-07-20
Refactor update config to use automatic retry of failed nodes for cluster fan out

ArgusLi:
## Description

Add setting button to monitor banner so that monitor settings can be configured mid run.

Add config update statistics to monitor banner.

Rework cluster fanout end-to-end so:

1. Node statuses will be classified into `attempting`, `retrying`, `succeeded`, `failed`, `not_attempted`.
    1. `attempting` is for nodes that are actively attempting.
    2. `retrying` is for nodes whose last attempt failed and are waiting out the backoff delay.
    3. `succeeded` is for nodes that have succeeded.
    4. `failed` is for nodes that have failed and have exhausted all retries.
    5. `not_attempted` is for nodes that have no resgistered metrics process and will not be retried.
2. The server will create a session for a cluster fanout action. All further actions will supercede an earlier action.
3. The server will fan out the action to all nodes within the cluster or the sole node in the case for standalone.
4. The server collects **per-node results**, with a retry mechanism following Fibonacci backoff, along with a total limit on number of attempts.
5. Per node statuses are streamed live to the frontend while the run is in progress.
6. When the session resolves, the per-node results are aggregated into a single reply (`sendConfigReply`). The reply is keyed by one `AggregateReplyId` arm (`clusterId` for a cluster, db-less `nodeId` for a standalone) and carries:
    1. an `outcome`: `fulfilled` when every attempted node succeeded, `failed` when any attempted node failed, `not_attempted` when no targeted node had a registered metrics process;
    2. the full `nodeResults` (one per attempted node) and the `notAttemptedNodeIds`;
    3. on `fulfilled` only, the echoed `appliedConfig`.
  On the wire this is just two message types: `updateConfigFulfilled` (sent only for `fulfilled`) and `updateConfigFailed` (sent for both `failed` and `not_attempted`).

## Changes
image
image
image


ravjotbrar:
How would multiple users sharing the same Valkey deployment be impacted with config aborts?

ArgusLi:
> How would multiple users sharing the same Valkey deployment be impacted with config aborts?

They won't. The config abort is purely for the in flight config update. If the config update is in flight (i.e. in the middle of running or waiting for a retry), sending the abort signal kills the config update completely preventing the action from finishing. It does not affect existing config or configs that have already been applied.

docs/adr/20260713-cluster-fan-out-retry-failed-nodes.md
```diff
@@ -0,0 +1,69 @@
+# ADR — Cluster fan out retry failed nodes
+
+- **Status:** Accepted
+- **Date:** 2026-07-13
+
+## Context
+
+Currently there are 3 cluster fan out actions: update metrics config, monitor start/stop, and enable cluster slot stats. All these fan out actions treat clusters as all-or-nothing. If a single node fails, the entire cluster action is considered failed.
```
ravjotbrar, line 8: Nit: Enabling cluster slot stats is actually not implemented. It was wired in, but it didn't make sense at the time to toggle Valkey settings if we weren't also polling them to get their most updated state.

docs/adr/20260713-cluster-fan-out-retry-failed-nodes.md
```diff
@@ -0,0 +1,69 @@
...
+2. The server will create a session for a cluster fanout action. All further actions will supercede an earlier action.
+3. The server will fan out the action to all nodes within the cluster or the sole node in the case for standalone.
+4. The server collects **per-node results**, with a retry mechanism following Fibonacci backoff, along with a total limit on number of attempts.
+5. Per node statuses are streamed live to the frontend while the run is in progress.
+6. When the session resolves, the per-node results are aggregated into a single reply (`sendConfigReply`). The reply is keyed by one `AggregateReplyId` arm (`clusterId` for a cluster, db-less `nodeId` for a standalone) and carries:
+  1. an `outcome`: `fulfilled` when every attempted node succeeded, `failed` when any attempted node failed, `not_attempted` when no targeted node had a registered metrics process;
+  2. the full `nodeResults` (one per attempted node) and the `notAttemptedNodeIds`;
+  3. on `fulfilled` only, the echoed `appliedConfig`.
+  On the wire this is just two message types: `updateConfigFulfilled` (sent only for `fulfilled`) and `updateConfigFailed` (sent for both `failed` and `not_attempted`).
```
ravjotbrar, line 30: Hmm I'm concerned that sending updateConfigFailed for not_attempted nodes would be confusing. Although, I haven't looked at how that's handled yet.

ravjotbrar: Scrolling further down I see that nodes that aren't attempted are still clear, so you can ignore this comment.

ravjotbrar:
> > How would multiple users sharing the same Valkey deployment be impacted with config aborts?
> 
> They won't. The config abort is purely for the in flight config update. If the config update is in flight (i.e. in the middle of running or waiting for a retry), sending the abort signal kills the config update completely preventing the action from finishing. It does not affect existing config or configs that have already been applied.

I'm more so referring to how User A's config update could be interrupted and superseded by User B's, but that's a rare enough collision case that I'm ok with that risk.

ravjotbrar (approved):
Looks great!

PR #393, opened 2026-07-21
upon refresh take the user back to the view they were in

### LINKED ISSUE #396: Persist authenticated session state across application refreshes
Currently, refreshing the page drops the WebSocket and all connection state, forcing the user back to the connections page. This disrupts workflow, especially in shared web deployments.

Keep users connected across page reloads using secure session restoration without storing credentials in plaintext. On refresh, the application should auto-reconnect to previously active connections rather than redirecting to the connection page.

### LINKED ISSUE #407: Persist command execution history across sessions
Users frequently re-run the same diagnostic commands. Currently, command history is lost on refresh, requiring retyping.

Save previously executed commands so users can view and reuse them across application sessions. History should persist across page refreshes and app restarts.

nassery318:
## Description

- redirect user to the view they were when they refreshed 
- show a spinner overlay while connecting

### Change Visualization

https://github.com/user-attachments/assets/f7e5759c-5068-4275-8283-5fecd499102c


common/src/constants.ts
```diff
@@ -137,6 +137,13 @@ export const LOCAL_STORAGE = {
   VALKEY_CONNECTIONS: "VALKEY_CONNECTIONS",
 }
 
+export const SESSION_STORAGE = {
+  VALKEY_COMMANDS: "VALKEY_COMMANDS",
+}
+
+// limit the number of commands persisted in session storage to avoid exceeding storage limits
+export const PERSISTED_COMMANDS_LIMIT = 25
```
ravjotbrar, line 145: I can see users wanting this to be configurable

apps/frontend/src/components/connection/Connection.tsx
```diff
@@ -28,6 +30,13 @@ export function Connection() {
   const [searchQuery, setSearchQuery] = useState("")
   const [passwordPromptConnectionId, setPasswordPromptConnectionId] = useState<string | undefined>(undefined)
   const connections = useSelector(selectConnections)
+  const [resumeWindowOpen, setResumeWindowOpen] = useState(true)
+
+  useEffect(() => {
+    // show the reconnecting overylay for short time
+    const timer = setTimeout(() => setResumeWindowOpen(false), 2000)
```
ravjotbrar: Is there a reason we go with a timeout here instead of connection state?

apps/frontend/src/history.ts
```diff
@@ -1,9 +1,17 @@
...
+
 const history = {
   location: null as unknown as Location,
   navigate: null as unknown as NavigateFunction,
   params: null as unknown as Readonly<Params<string>>,
+  refreshedFrom: (refreshPath.startsWith("/") ? refreshPath : null) as string | null,
 }
 
+export const wasRefreshedFrom = (connectionId: string): boolean =>
+  Boolean(history.refreshedFrom?.includes(`/${connectionId}/`)) &&
```
ravjotbrar: Segments might be more robust here or using params:

```
    const segments = history.refreshedFrom.split("/")
    return segments.includes(connectionId) &&
      !history.refreshedFrom.endsWith("/valkey-reconnect")
```

apps/frontend/src/history.ts
```diff
@@ -1,9 +1,17 @@
 import type { NavigateFunction, Params, Location } from "react-router"
 
+// the path user refershed from
```
ArgusLi: nit: refreshed

apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -433,6 +434,21 @@ export const sendRequestEpic = () =>
     }),
   )
 
+export const persistCommandsEpic = (store: Store) =>
+  action$.pipe(
+    selectMany(sendFulfilled, sendFailed, deleteConnection),
```
ravjotbrar: Could you clarify why we filter for deleteConnection? It seems like if we wanted a side effect when a connection is deleted, that side effect should go into the deleteConnectionEpic.

apps/frontend/src/history.ts
nassery318: good catch

apps/frontend/src/state/epics/valkeyEpics.ts
nassery318: persistCommandsEpic is the only place that writes VALKEY_COMMANDS, so it listens for anything that changes command state. Deleting a connection clears the connection command history in the reducer, and if we don't re-persist, it comes back on refresh. So doing it in deleteConnectionEpic would mean writing VALKEY_COMMANDS from two places.

common/src/constants.ts
nassery318: okay this will require config modal just like the big/hot keys. should I add it in this PR?

apps/frontend/src/history.ts
nassery318: using segments for robustness and better readability

apps/frontend/src/components/connection/Connection.tsx
nassery318: used the RECONNECTING state instead of the arbitrary timeout.

common/src/constants.ts
nassery318: add config capability

apps/frontend/src/state/valkey-features/command/commandSlice.ts
```diff
@@ -28,25 +46,49 @@ const withMetadata = (command: string, response: JSONObject, isFulfilled = true)
...
       return {
         ...state,
-        [connectionId]: {
-          pending: false,
-          commands: [cmd, ...prev],
+        connections: {
+          ...state.connections,
+          [connectionId]: {
+            pending: false,
+            commands: [cmd, ...prev],
```
ravjotbrar, line 90: I think we need .slice(0, state.limit) so the sessionStorage doesn't grow unbounded, like in your previous commit. Same goes for sendFailed below.

apps/frontend/src/components/connection/Connection.tsx
```diff
@@ -55,6 +57,13 @@ export function Connection() {
     }))
   }
 
+  const isResuming = Object.entries(connections).some(([connectionId, connection]) =>
+    wasRefreshedFrom(connectionId) &&
+    (connection.status === RECONNECTING ||
+      connection.status === CONNECTING ||
+      connection.status === CONNECTED),
```
ravjotbrar, line 66: Did we need CONNECTED here?

nassery318: Yes, if we drop the CONNECTED we will see a flash of the connected list prior to getting redirected. Having it will make it close to seamless.

nassery318: can add a comment to clarify it in the code.

apps/frontend/src/state/valkey-features/command/commandSlice.ts
nassery318: sessionStorage is already capped, the persistCommandsEpic slices to limit before saving, so it won't grow unbounded. The in-memory list is kept full on purpose so raising the limit can show older commands again.

ravjotbrar (approved):

PR #410, opened 2026-07-22
Fix Electron CSP blocking websocket connections

ravjotbrar:
## Description
 The onHeadersReceived CSP in electron.main.js had connect-src 'self' which overrides the permissive meta tag in index.html. When both are present, the browser enforces the intersection (most restrictive). Add ws://localhost:* and http://localhost:* to match the meta tag.

### Change Visualization

### DIFF
--- a/apps/frontend/electron.main.js
+++ b/apps/frontend/electron.main.js
@@ -65,7 +65,7 @@ app.whenReady().then(() => {
         "Content-Security-Policy": [
           "default-src 'self'; script-src 'self'; " +
           "style-src 'self' 'unsafe-inline'; " +
-          "connect-src 'self'; " +
+          "connect-src 'self' ws://localhost:* http://localhost:*; " +
           "img-src 'self' data:; font-src 'self'; " +
           "object-src 'none'; base-uri 'self'; form-action 'self';",
         ],


apps/frontend/electron.main.js
```diff
@@ -65,7 +65,7 @@ app.whenReady().then(() => {
         "Content-Security-Policy": [
           "default-src 'self'; script-src 'self'; " +
           "style-src 'self' 'unsafe-inline'; " +
-          "connect-src 'self'; " +
+          "connect-src 'self' ws://localhost:* http://localhost:*; " +
```
alexey-temnikov, line 68: Is there the ability to scope to specific ports, not all?

alexey-temnikov, line 68: It seems like we need to keep it consistent. - https://github.com/valkey-io/valkey-admin/blob/main/apps/frontend/index.html . Would it be better to add a small comment cross-referencing them?

alexey-temnikov (approved):
Approved with few questions.

apps/frontend/electron.main.js
ravjotbrar: Unfortunately not, due to the user having the ability to override the main server's default port, which is 8080.

ravjotbrar: We could probably actually remove this one and keep the meta tag in index.html. Will do in a future PR to avoid any big changes before the release.

PR #426, opened 2026-07-28
Make search case-insensitive in Send Command view

### LINKED ISSUE #424: History and response search in Send Command should be case-insensitive
The search bars in the Send Command view for both History and Response only return matches when the search term uses the exact same letter casing as the stored text.

For example:

Searching for **HELLO** will find **HELLO** commands.
Searching for **hello** will not return the same results.

The search behaviour should be case-insensitive so users can find matching commands and responses regardless of whether they enter uppercase, lowercase, or mixed-case search terms.

antonin-suzor:
## Description

Closes #424 

This PR makes search case-insensitive in the Send Command view.
Affected search bars are:
- History search bar
- Response search bar
- Diff search bar

This works by lowercasing the filter (once per change/render) and lowercasing the entries (once per entry).

### Known limitation

Text highlighting in diff view is still case-sensitive in `CommandElements.tsx`/`KeyFilterable`.

I did not find a straight-forward way to implement it, and did not know if it was included in the scope of the issue.

A possible implementation (not explored) would be to manually split the "normal-case" keyPathString based on split results of the lowercase filter against a lowercased keyPathString. If it should be included in the PR, I can look to add it.

### Tests ran

- Manual testing of the feature
- `npm run build:frontend`
- `npm run lint`

### Change Visualization

Before the changes:
https://github.com/user-attachments/assets/855c386f-39e1-4ecc-91c9-7d56e7698cac

After the changes:
https://github.com/user-attachments/assets/8873c25f-8f08-4230-be43-168140f2f907

### DIFF
--- a/apps/frontend/src/components/send-command/DiffCommands.tsx
+++ b/apps/frontend/src/components/send-command/DiffCommands.tsx
@@ -16,7 +16,8 @@ const DiffCommands = ({ filter, id, indexA, indexB }) => {
 
   if (R.isEmpty(diffs)) return "Responses are identical"
 
-  const filteredDiffs = filter ? diffs.filter(({ keyPathString }) => keyPathString.includes(filter)) : diffs
+  const normalisedFilter = filter.toLowerCase()
+  const filteredDiffs = normalisedFilter ? diffs.filter(({ keyPathString }) => keyPathString.toLowerCase().includes(normalisedFilter)) : diffs
 
   const onCopy = () =>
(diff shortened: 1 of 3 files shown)


ravjotbrar (approved):
Thanks for the PR @antonin-suzor. LGTM!

antonin-suzor:
Thank you for the review !

I don't have merge permission, so you will have to merge it yourself @ravjotbrar

PR #434, opened 2026-07-30
Fix hot keys returning empty on first MONITOR cycle

### LINKED ISSUE #433: Hot keys returns empty on first MONITOR cycle due to disk write race
When starting MONITOR for the first time (no prior data on disk), the hot keys fetch returns empty results even after the full monitoring duration completes. Clicking refresh immediately after shows the correct hot keys.

#### Root Cause:
  
The server waits until checkAt (set to startTime + monitoringDuration) then re-fetches hot keys from the metrics process. However, checkAt marks when collection ends in memory, not when the data is written to disk asynchronously after collectLogs returns via await onLogs(rows). The re-fetch races the disk write and reads an empty file.
  
#### Fix:
Add a 1-second buffer to the server-side delay after checkAt. The actual write overhead is ~50ms for typical cycles, so 1 second provides 20x margin.

ravjotbrar:
### Summary
  
Prior to this change, the server waits until checkAt (collection end time) then immediately re-fetches /hot-keys. But monitor data is collected in memory for the full duration, then written to disk after collectLogs returns. The re-fetch arrives before the write completes, returning empty results.
  
This PR adds a 1-second buffer to the hot keys re-fetch delay after checkAt to account for the asynchronous disk write that happens after monitor collection ends. The 1-second buffer accounts for: promise resolution + event listener cleanup + JSON serialization of rows + fs.appendFile. 

### Before the fix

https://github.com/user-attachments/assets/9473b28c-e5e5-429a-99a4-b7b5171e3890

### After the fix

https://github.com/user-attachments/assets/c89589cd-be35-4966-86e9-aadbef252661

### DIFF
--- a/apps/server/src/actions/hotkeys.ts
+++ b/apps/server/src/actions/hotkeys.ts
@@ -84,19 +84,19 @@ export const hotKeysRequested = withDeps<Deps, void>(
         if (initialParsedResponse.checkAt) {
           // Monitor collects data in memory for the full duration, then writes
           // to disk asynchronously after collectLogs returns. The re-fetch may
-          // arrive before the write completes. Wait checkAt + 1s, then retry up
-          // to 3 times (1s apart) if data still isn't available. 
-          const delay = initialParsedResponse.checkAt - Date.now() + 1000
+          // arrive before the write completes. Wait until checkAt, then retry up
+          // to 3 times (1s apart) if data still isn't available.
+          const delay = initialParsedResponse.checkAt - Date.now()
           await new Promise((resolve) => setTimeout(resolve, Math.max(delay, 0)))
 
+          let result: HotKeysResponse = initialParsedResponse
           for (let attempt = 0; attempt < 3; attempt++) {
-            const dataResponse = await fetch(url)
-            const result = await dataResponse.json() as HotKeysResponse
-            if (result.hotKeys?.length > 0 || !result.monitorRunning) return result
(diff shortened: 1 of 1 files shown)


apps/server/src/actions/hotkeys.ts
```diff
@@ -82,7 +82,10 @@ export const hotKeysRequested = withDeps<Deps, void>(
         const initialParsedResponse: HotKeysResponse = await initialResponse.json() as HotKeysResponse
         // Reads monitor data and returns when to fetch results (`checkAt`).
         if (initialParsedResponse.checkAt) {
-          const delay = initialParsedResponse.checkAt - Date.now()
+          // Add 1s buffer: checkAt marks when collection ends, but data is written
```
nassery318: wondering, what if the disk takes longer than 1s (it is under load, or it is a slow disk)? the refetch will still return empty.

ravjotbrar: Good question. Initially, I thought 1s delay is a good tradeoff. Now that I think about it maybe we should do a limit of retries until we hit that limit or until we get data back.

nassery318: good point

ravjotbrar: Made the changes

ArgusLi (approved):
The "best" solution would somehow be to send a signal after disk write, then we wait for that before getting the result, but that would be completely overkill. This works.

PR #436, opened 2026-07-30
Fix monitor and node error banners

### LINKED ISSUE #425: START MONITOR warning modal does not handle partial cluster connections correctly
The **START MONITOR** warning modal and its configuration settings do not behave as expected when working with a cluster that has multiple nodes.

When starting the monitor, if the application fails to monitor only one of the cluster nodes (while successfully monitoring the others), the warning modal when stopped does not close while the stop button and setting disappears.

Additionally, changes made to the Hot Keys configuration from the START MONITOR warning modal are not applied correctly. The configuration updates appear to have no effect after being saved.

**Steps to Reproduce**
1. Connect to a cluster with multiple nodes.
2. Ensure one of the nodes fails to connect while the remaining nodes connect successfully.
3. Open START MONITOR.
4. Observe that the warning modal does not dismiss or behave as expected.
5. Change the Hot Keys configuration from the warning modal settings.
6. Save the changes and verify the updated configuration.

Image
Image

ArgusLi:
## Description
Addresses #425.

Instead of reading the current node's entry, which made the error invisible if a different node failed, monitorError is derived from the first cluster node whose MONITOR failed to start.

For the Hot Keys config changes not being reflected, the sendConfigReply now includes appliedConfig in the partial reply, so that the frontend will update config even on partial successes.

### Change Visualization

image

### DIFF
--- a/apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
+++ b/apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
@@ -62,7 +62,9 @@ export function HotKeys({
 
   const banners = (
     <>
-      {!monitorRunning && onStartMonitoring && (
+      {/* Also shown while running when a node's monitor failed to start
+          (partial cluster): the error variant reports the failed node. */}
+      {(!monitorRunning || monitorError) && onStartMonitoring && (
         <MonitorNotRunningBanner error={monitorError} onStartMonitoring={onStartMonitoring} />
       )}
       {nodeErrors && nodeErrors.length > 0 && (
--- a/apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
+++ b/apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
@@ -62,9 +62,7 @@ export function HotKeys({
 
   const banners = (
     <>
-      {/* Also shown while running when a node's monitor failed to start
-          (partial cluster): the error variant reports the failed node. */}
-      {(!monitorRunning || monitorError) && onStartMonitoring && (
+      {!monitorRunning && onStartMonitoring && (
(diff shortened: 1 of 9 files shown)


apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
```diff
@@ -62,7 +62,9 @@ export function HotKeys({
 
   const banners = (
     <>
-      {!monitorRunning && onStartMonitoring && (
+      {/* Also shown while running when a node's monitor failed to start
+          (partial cluster): the error variant reports the failed node. */}
+      {(!monitorRunning || monitorError) && onStartMonitoring && (
```
ravjotbrar: I believe this needs to be reverted

apps/frontend/src/state/valkey-features/monitor/monitorSlice.ts
```diff
@@ -19,6 +19,20 @@ export const selectMonitorError =
     (state: RootState) =>
       R.path<string | null>([VALKEY.MONITOR.name, connectionId, "error"], state) ?? null
 
+/**
+ * Per-node monitor errors for a cluster (e.g. a node whose MONITOR failed to
+ * start because it is unreachable), so the UI can report a partially-running
+ * monitor instead of silently dropping the failed nodes.
+ */
+export const selectClusterMonitorNodeErrors =
```
ravjotbrar, line 27: As a result of the above comment, we could probably remove this selector and the code that uses it too

apps/frontend/src/components/activity-view/hotkeys/hot-keys.tsx
nassery318: yes,  the NodeErrorsBanner already covers/display the single or multiple nodes Error.

nassery318 (approved):

PR #451, opened 2026-08-05
Pipeline per-key commands for performance

### LINKED ISSUE #450: Big Keys scan is slow due to sequential per-key commands
## Description

The Big Keys scan issues 3 separate commands per key (`MEMORY USAGE`, `TYPE`, `TTL`) sequentially for every sampled key. With the default scan limit of 10K keys, this results in 30,000 individual round trips to Valkey over TLS, taking ~40 seconds on a 25-shard ElastiCache cluster.

At higher scan limits (100K+), the volume of in-flight commands can crash metrics server processes.

## Root Cause

`apps/metrics/src/analyzers/scan-big-keys.js` processes each key individually within a SCAN batch:

```js
for (const key of keys) {
  const [sizeBytes, type, ttl] = await Promise.all([
    client.customCommand(["MEMORY", "USAGE", key, "SAMPLES", "5"]),
    client.customCommand(["TYPE", key]),
    client.customCommand(["TTL", key]),
  ])
}
```

## Fix

Pipeline all per-key commands into a single `Batch` (non-atomic pipeline) per SCAN iteration, reducing round trips from 3×N to 1 per batch.

ravjotbrar:
Performance enhancement that pipelines per-key Valkey commands instead of issuing them as individual round trips. Applied across three areas:

1. **Big Keys scan**:`MEMORY USAGE`, `TYPE`, `TTL`, and `OBJECT FREQ` are now pipelined per SCAN batch instead of sequentially per key. Default batch size increased from 100 to 500.

2. **Hot Keys enrichment**: `TTL` and `MEMORY USAGE` for hot key results are pipelined in a single batch instead of per-key `Promise.all` calls.

3. **Key Browser**: Metadata fetch (`TYPE`, `TTL`, `MEMORY USAGE`) for all scanned keys is pipelined upfront via `fetchKeyMetadataBatch`, then passed into `getKeyInfo` for type-specific follow-ups.

All pipelines are capped at 500 keys per batch to avoid overwhelming the server.

Closes #450

### DIFF
--- a/apps/metrics/src/analyzers/calculate-hot-keys.js
+++ b/apps/metrics/src/analyzers/calculate-hot-keys.js
@@ -76,6 +76,8 @@ export const calculateHotKeysFromMonitor = ({ limit, cutoff }) => (rows) =>
     R.take(limit),
   )(rows)
 
+import { Batch } from "@valkey/valkey-glide"
+
 // Must have maxmemory-policy set to lfu*
 export const calculateHotKeysFromHotSlots = async (client, { count = 50 } = {}) => {
   const hotSlots = await getHotSlots(client)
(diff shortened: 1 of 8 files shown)


apps/metrics/src/analyzers/calculate-hot-keys.js
```diff
@@ -93,26 +94,34 @@ export const calculateHotKeysFromHotSlots = async (client, { count = 50 } = {})
...
   })
 
-  const keyFreqNestedPromises = await Promise.all(slotPromises)
-  const keyFreqPromises = keyFreqNestedPromises.flat()
-  const allKeyFreqs = await Promise.all(keyFreqPromises)
+  const slotKeys = await Promise.all(slotPromises)
+  const allKeys = slotKeys.flat()
+
+  // Pipeline OBJECT FREQ in chunks to avoid overwhelming the server
+  const PIPELINE_CHUNK_SIZE = 500
```
ArgusLi: PIPELINE_CHUNK_SIZE should be moved to constants. It's referred to in calculate-hot-keys, enrich-hot-keys and keys-browser.

ravjotbrar: Good shout

apps/metrics/src/analyzers/scan-big-keys.js
```diff
@@ -37,14 +43,23 @@ export const scanBigKeys = async (client, { scanLimit = 10000, topN = 50, batchS
...
+
+  let freqResults
+  try {
+    freqResults = await client.exec(freqBatch)
+  } catch {
+    freqResults = null
+  }
+
+  const keys = topKeys.map((entry, i) => {
+    const freq = freqResults?.[i] != null ? Number(freqResults[i]) : null
```
alexey-temnikov: Have you tested this for OBJECT FREQ when LFU is not enabled?

ravjotbrar: I tested and we actually hide the frequency column when results from OBECT FREQ are null. It turns out JSON.stringify(Nan) -> null so it's hidden from the user. That being said we should still fix the issue and I've added a commit that normalizes error responses from OBJECT FREQ

apps/metrics/src/analyzers/scan-big-keys.js
```diff
@@ -1,6 +1,7 @@
 import { Heap } from "heap-js"
+import { Batch } from "@valkey/valkey-glide"
 
-export const scanBigKeys = async (client, { scanLimit = 10000, topN = 50, batchSize = 100 } = {}) => {
+export const scanBigKeys = async (client, { scanLimit = 10000, topN = 50, batchSize = 500 } = {}) => {
```
nassery318: might wanna use the PIPELINE_CHUNK_SIZE instead of 500

ravjotbrar: Good catch

alexey-temnikov (approved):

Issue #438, opened 2026-08-04
Replace database dropdown with number input to support configurable database counts


ravjotbrar:
### Description

The connection modal currently uses a hardcoded dropdown with 16 entries (DB 0–15) for database selection. Users who configure their Valkey server with more than 16 databases (via the databases parameter) cannot connect to databases beyond 15.
  
### Proposed Change
Replace the database dropdown with a NumberInput (step counter) component. The input should have min=0 and step=1. The first time a user connects, no max can be set as we don't have a means to dynamically infer what they set `databases` to.

PR #452, opened 2026-08-05
Update connections to allow numbered databases beyond 16

ArgusLi:
## Description

Addresses #438.

Previously, numbered databases were locked to the default 0-15.

This PR:
- Replaces the dropdown with a number input that doesn't limit database id to 15.
- Adds an error message when the connection database id is outside of the range set in the Valkey instance.
- When editing an existing connection, shows what the acceptable numbered database range is.
- Fixes a bug where a raw glide error is seen by users for out-of-range db id.

### Change Visualization

image

image

image


docs-site/src/content/docs/reference/troubleshooting.md
```diff
@@ -57,14 +57,30 @@ The server rejected the connection because the `db` field on the connection payl
...
+The supplied `db` is at or beyond the target server's configured database count. Valkey's `databases` config defaults to `16` (giving `0..15`), but operators can lower or raise it. The server reads this value via `CONFIG GET databases` on the probe client before binding the requested database, so the bound depends on what the server is actually configured for, not a fixed constant.
 
 **Fix:** Pick a `db` within `[0, databases - 1]`. To check or change the server's bound, run `CONFIG GET databases` against the target node, and `CONFIG SET databases <N>` (or update `valkey.conf` / `redis.conf` and restart) if you need more logical databases. Note that ElastiCache and other managed services may not allow changing `databases`, in which case you must stay within the operator-set limit.
 
 ---
 
+## Connection rejected: `Database_Index N is out of range: this cluster has cluster-databases = M (valid range 0..M-1)`
+
+The cluster variant of the error above. Valkey 9.0+ clusters configure their logical database count via `cluster-databases` (falling back to `databases` when unset), and the requested `db` is at or beyond that count.
```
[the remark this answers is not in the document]
ArgusLi: The app falls back to databases.

apps/server/src/connection.ts
```diff
@@ -213,13 +231,18 @@ async function connectToValkeyLocked(
...
         checkJsonModuleAvailability(existingStandalone),
+        getDatabasesCount(existingStandalone),
       ])
       sendStandaloneConnectFulfilled(ws, {
         connectionId,
-        connectionDetails: { keyEvictionPolicy, jsonModuleAvailable },
+        connectionDetails: {
+          keyEvictionPolicy,
+          jsonModuleAvailable,
+          databasesCount: existingDatabasesCount ?? DEFAULT_DATABASES_COUNT,
```
ravjotbrar: I think a default doesn't work, since the user could still have more than the default count even if we weren't able to learn it through the config command.

ArgusLi: Yup, removing as we speak.

ravjotbrar (approved):

PR #471, opened 2026-08-12
Use subpath in websocket URL

tkesgar:
Slightly adjusts the websocket URL to use subpath. This allows us to support subpath deployments without having to rebuild the frontend:

- Normal deployment:
  - URL: `https://valkey.mywebsite.com/`
  - WS URL: `wss://valkey.mywebsite.com/`
- Subpath deployment:
  - URL: `https://mywebsite.com/valkey/`
  - WS URL: `wss://mywebsite.com/valkey/`

Currently we use `ws[s]://window.location.host`, which ignores the root path of our subpath deployment even with rewrite in reverse proxy (e.g. `https://mywebsite.com/valkey/` => `wss://https://mywebsite.com`).

We can instead derive the websocket URL from `window.location.origin + window.location.pathname`, which gives us the full path without hash part or search query.

### DIFF
--- a/apps/frontend/src/state/epics/wsEpics.ts
+++ b/apps/frontend/src/state/epics/wsEpics.ts
@@ -27,13 +27,13 @@ import {
 let socket$: WebSocketSubject<PayloadAction> | null = null
 
 const getWebsocketURL = () => {
-  const isElectron = window.location.protocol === "file:"
-  if (isElectron) {
+  // If it's an Electron deployment
+  if (window.location.protocol === "file:") {
     return "ws://localhost:8080"
   }
   
-  return (window.location.origin + window.location.pathname)
-    .replace(/^http/, 'ws')
+  const protocol = window.location.protocol === "https:" ? "wss" : "ws"
+  return `${protocol}://${window.location.host}${window.location.pathname}`
 }
 
 const url =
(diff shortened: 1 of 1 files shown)


apps/frontend/src/state/epics/wsEpics.ts
```diff
@@ -26,14 +26,19 @@ import {
...
-const isHttps = window.location.protocol === "https:"
+const getWebsocketURL = () => {
+  const isElectron = window.location.protocol === "file:"
+  if (isElectron) {
+    return "ws://localhost:8080"
+  }
+  
+  return (window.location.origin + window.location.pathname)
+    .replace(/^http/, 'ws')
+}
```
ravjotbrar, line 37: Instead of using the string replacement trick, I'd recommend just appending the pathname:

```suggestion
const getWebsocketURL = () => {
  // If it's an Electron deployment
  if (window.location.protocol === "file:") {
    return "ws://localhost:8080"
  }
  
  const protocol = window.location.protocol === "https:" ? "wss" : "ws"
  return `${protocol}://${window.location.host}${window.location.pathname}`
}
```

Otherwise the change is reasonable!

ravjotbrar (approved):

ravjotbrar:
One thing to note: users deploying behind a subpath proxy will need to set VALKEY_ADMIN_ALLOWED_WS_ORIGINS to include the proxy's origin (e.g., http://myproxy:9090)

PR #481, opened 2026-08-15
Add scan button for Big Keys, revamp Start Monitor button

### LINKED ISSUE #447: Add scan button for Big Keys and revamp Start Monitor button for Hot Keys
## Description

Two UX improvements for the Activity view:

1. **Big Keys**: Add a more prominent scan/re-scan button. Currently the only way to re-trigger a scan is via the generic refresh button or the params modal. A dedicated "Scan" button would make the action more discoverable.

2. **Hot Keys**: Revamp the "Start Monitor" button to look more like a button and less like a link.

ravjotbrar:
Closes #447

- Big Keys: replaced the icon-only refresh button with a prominent, labeled "Scan" button
- Hot Keys: replaced the underlined text-link "Start MONITOR" with an actual Button component (both instances — the banner and the empty-state variant)

Also added/fixed while in this area:
- Cluster connection group header now shows an "Open" button when connected (matching standalone connections), navigating to the cluster topology view
- Individual node connections within an expanded cluster group also get an "Open" button, navigating to that node's dashboard
- Redesigned `TabGroup` (Hot Keys / Big Keys / Command Logs) as underline-style tabs instead of button pills — the previous pill style made action buttons using the same `default` variant (like Start Monitor) look like an extra tab
- Removed a duplicate "Start Monitor" button that rendered twice (banner + empty state) when Hot Keys had no data and monitor wasn't running 

Screenshot 2026-08-14 at 5 53 46 PM

Screenshot 2026-08-14 at 5 54 26 PM

Screenshot 2026-08-14 at 5 52 37 PM

### DIFF
--- a/apps/frontend/src/components/activity-view/ActivityView.tsx
+++ b/apps/frontend/src/components/activity-view/ActivityView.tsx
@@ -257,8 +257,9 @@ export const ActivityView = () => {
                 <Settings className="hover:text-primary" size={15} />
               </Button>
             )}
-            {!useHotSlots && !monitorRunning && (
+            {!useHotSlots && (
               <Button
+                disabled={monitorRunning}
                 onClick={() => setConfigOpen(true)}
                 size={"sm"}
                 variant={"default"}
--- a/apps/frontend/src/components/activity-view/ActivityView.tsx
+++ b/apps/frontend/src/components/activity-view/ActivityView.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react"
 import { useSelector } from "react-redux"
-import { Activity, RefreshCcw, ScanSearch, Settings } from "lucide-react"
+import { Activity, Play, RefreshCcw, ScanSearch, Settings } from "lucide-react"
 import { useParams } from "react-router"
 import { COMMANDLOG_TYPE, PENDING } from "@common/src/constants"
 import { truncateText } from "@common/src/truncate-text"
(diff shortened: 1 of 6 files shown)


ArgusLi:
Should we move the start monitor button directly out of the banner? In terms of UX design, doesn't it make sense to separate the warning banner from the action? I think placing the button on the right side of the tab group, like the Big Keys scan button, would make more sense.

ravjotbrar:
> Should we move the start monitor button directly out of the banner? In terms of UX design, doesn't it make sense to separate the warning banner from the action? I think placing the button on the right side of the tab group, like the Big Keys scan button, would make more sense.

I wanted to make the solution to monitor not running as obvious as possible, but I agree with you that it looks cleaner on the right side similar to big keys scan.

apps/frontend/src/components/activity-view/ActivityView.tsx
```diff
@@ -264,6 +264,16 @@ export const ActivityView = () => {
             >
               <RefreshCcw className="hover:text-primary" size={15} />
             </Button>
+            {!useHotSlots && !monitorRunning && (
```
ArgusLi: Instead of having the button disappear if either useHotSlots or monitorRunning. Would it be better if button only disappears if useHotSlots. Then, it gets disabled if monitorRunning?

ravjotbrar: Done

ArgusLi (approved):

nassery318 (approved):

PR #486, opened 2026-08-19
first implementatoin of cluster node aggregation view

### LINKED ISSUE #395: Add cluster node aggregation view with side-by-side comparison
Currently, users must navigate to individual nodes to understand cluster health. A unified view enables faster diagnosis of hotspots, memory imbalance, and performance drift across the cluster.

We need to add a consolidated cluster view that shows aggregated metrics (total memory/capacity, total throughput, cluster-wide hit ratio) alongside per-node breakdowns. Enable side-by-side comparison to identify imbalance, drift, or outlier behavior between nodes and shards.

Side by side comparison can be done by merging this new functionality with the cluster topology route.

nassery318:
## Description

Merging cluster-wide metrics into the existing Cluster Topology route, so topology and utilization are one screen instead of visiting each node.

- Add a stat card row: Total Nodes, Cluster Memory, Total Ops/Sec, Cluster Hit Ratio, and Nodes Flagged.
- Replace the nested primary/replica list with a flat table — one row per node, showing role, utilization, memory, CPU, ops/sec, hit ratio and connections.
- Add a utilization badge (Low / Normal / High) per primary, based on whichever of memory or CPU is worse.
- Add role and utilization filters alongside the existing search.
- Compute per-node utilization on the server from the existing INFO fan-out, and send it with the cluster payload.
- Derive CPU as a live rate by diffing the main-thread CPU counters between polls, falling back to process-wide counters where those are not reported.
- Measure memory against `maxmemory`, falling back to host RAM when no limit is set (shown with a dashed badge). - Poll cluster data every 5s while the route is mounted, and stop on unmount.

### Change Visualization

BEFORE:

image

AFTER:

image

### DIFF
--- a/apps/frontend/src/components/cluster-topology/Cluster.tsx
+++ b/apps/frontend/src/components/cluster-topology/Cluster.tsx
@@ -21,10 +21,10 @@ import {
   selectCluster, selectClusterNodeRows, selectClusterMetrics
 } from "@/state/valkey-features/cluster/clusterSelectors"
 import { useAppDispatch } from "@/hooks/hooks"
-import { updateClusterData, stopClusterDataPolling } from "@/state/valkey-features/cluster/clusterSlice"
+import { updateClusterData, stopClusterDataPolling, type NodeRole } from "@/state/valkey-features/cluster/clusterSlice"
 import { selectClusterAlias } from "@/state/valkey-features/connection/connectionSelectors"
 
-type RoleFilter = "all" | "primary" | "replica"
+type RoleFilter = "all" | NodeRole
 type UtilizationFilter = "all" | UtilizationLevel
 
 export function Cluster() {
@@ -179,7 +179,7 @@ export function Cluster() {
               key={row.dataKey}
               nodeData={clusterData.data[row.dataKey]}
               port={row.port}
-              primary={row.primary}
+              primaryConfig={row.primary}
               role={row.role}
               utilization={clusterData.utilization?.[row.dataKey]}
             />
--- a/apps/frontend/src/components/cluster-topology/Cluster.tsx
+++ b/apps/frontend/src/components/cluster-topology/Cluster.tsx
@@ -6,7 +6,6 @@ import { MAX_CONNECTIONS } from "@common/src/constants.ts"
 import { truncateText } from "@common/src/truncate-text.ts"
 import { formatBytes } from "@common/src/bytes-conversion.ts"
 import { calculateHitRatio } from "@common/src/cache-hit-ratio.ts"
-import { sanitizeUrl } from "@common/src/url-utils.ts"
 import { AppHeader } from "../ui/app-header"
 import RouteContainer from "../ui/route-container"
 import { StatCard } from "../ui/stat-card"
@@ -16,25 +15,17 @@ import { Typography } from "../ui/typography"
 import { TableContainer } from "../ui/table-container"
 import { StaticTableHeader } from "../ui/sortable-table-header"
 import { ClusterNodeRow } from "./cluster-node-row"
-import { getUtilizationLevel, type UtilizationLevel } from "./node-metrics"
-import type { PrimaryNode } from "@/state/valkey-features/cluster/clusterSlice"
-import { selectCluster } from "@/state/valkey-features/cluster/clusterSelectors"
+import type { RootState } from "@/store.ts"
+import { getUtilizationLevel, type UtilizationLevel } from "@/state/valkey-features/cluster/clusterUtilization"
+import {
+  selectCluster, selectClusterNodeRows, selectClusterMetrics
+} from "@/state/valkey-features/cluster/clusterSelectors"
 import { useAppDispatch } from "@/hooks/hooks"
 import { updateClusterData, stopClusterDataPolling } from "@/state/valkey-features/cluster/clusterSlice"
 import { selectClusterAlias } from "@/state/valkey-features/connection/connectionSelectors"
 
 type RoleFilter = "all" | "primary" | "replica"
 type UtilizationFilter = "all" | UtilizationLevel
-
-interface NodeRow {
-  dataKey: string
-  searchKey: string
-  host: string
-  port: number
-  role: "primary" | "replica"
-  primary: PrimaryNode
-  isGroupEnd: boolean
-}
 
 export function Cluster() {
   const { id, clusterId } = useParams()
@@ -46,6 +37,8 @@ export function Cluster() {
     }
   }, [id, clusterId, dispatch])
   const clusterData = useSelector(selectCluster(clusterId!))
+  const nodeRows = useSelector((state: RootState) => selectClusterNodeRows(state, clusterId!))
+  const metrics = useSelector((state: RootState) => selectClusterMetrics(state, clusterId!))
   const [searchQuery, setSearchQuery] = useState("")
   const [roleFilter, setRoleFilter] = useState<RoleFilter>("all")
   const [utilizationFilter, setUtilizationFilter] = useState<UtilizationFilter>("all")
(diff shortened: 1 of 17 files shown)


apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
```diff
@@ -116,18 +135,20 @@ export function ClusterNode({
...
-        <MemoryStick className="text-primary" size={14} />
-        <Typography variant="bodyXs">{nodeData?.used_memory_human ?? "N/A"}</Typography>
-      </div>
-      <div className="flex items-center gap-1">
-        <Users className="text-primary" size={14} />
-        <Typography variant="bodyXs">{nodeData?.connected_clients ?? "N/A"}</Typography>
-      </div>
-    </div>
-  )
+  const usedMemory = utilization?.used_memory ?? (Number(nodeData?.used_memory) || 0)
```
ravjotbrar: Why do we default to 0?

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
```diff
@@ -116,18 +135,20 @@ export function ClusterNode({
...
+  const memoryLimit = utilization?.memory_limit_bytes ?? 0
```
ravjotbrar: Same question as above

apps/server/src/connection.ts
```diff
@@ -742,6 +743,12 @@ export function teardownConnection(
   const connection = clients.get(connectionId)
   clients.delete(connectionId)
 
+  // Samples are keyed per node, so keep them while any other db shares this node.
+  const nodeId = toNodeId(connectionId)
+  if (![...clients.keys()].some((id) => toNodeId(id) === nodeId)) {
```
ravjotbrar: This seems like a redundant check. Line 752 already checks if any other connections exist that share the same client as the current connection. All connectionIds that share the same nodeId will also share the same client.

apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -91,60 +151,94 @@ export function Cluster() {
...
-          icon={<CheckCircle2 className="text-green-500" size={20} />}
-          label="Connected"
-          value={connectedNodes}
-        />
+      <div className="grid grid-cols-2 md:grid-cols-3 lg:grid-cols-5 gap-4">
+        <StatCard label="Total Nodes" value={nodeRows.length} />
+        <StatCard label="Cluster Memory" value={clusterMemoryValue} />
+        <StatCard label="Total Ops/Sec" value={totalOpsPerSecValue} />
+        <StatCard label="Cluster Hit Ratio" value={clusterHitRatioValue} />
+        <StatCard label="Nodes Flagged" value={flaggedNodes} />
```
ravjotbrar: We can probably combine total nodes and nodes flagged by showing nodes flagged / total nodes in one stat card.

apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -50,33 +67,76 @@ export function Cluster() {
...
-  // cluster stats
-  const totalNodes = clusterEntries.length
-  const connectedNodes = clusterEntries.filter(([, primary]) =>
-    connectionStatuses[buildConnectionId(primary.host, primary.port, clusterDb)]?.status === CONNECTED,
-  ).length
-  const totalReplicas = clusterEntries.reduce((sum, [, primary]) =>
-    sum + primary.replicas.length, 0,
-  )
-  const totalClusterNodes = totalNodes + totalReplicas
+  const nodeRows: NodeRow[] = clusterEntries.flatMap(([primaryKey, primary]) => [
```
ravjotbrar: Our components should focus on rendering prepared state. We should build nodeRows in a selector.

apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -50,33 +67,76 @@ export function Cluster() {
...
+  let memoryLimitSum = 0
+  let opsPerSecSum = 0
+  let hitsSum = 0
+  let missesSum = 0
+  let flaggedNodes = 0
 
-    // check primary node
-    if (clusterData.searchableText[primaryKey]?.includes(searchQuery)) {
-      return true
+  for (const row of nodeRows) {
```
ravjotbrar: This should also be moved to a selector. We also want to be able to unit test without rendering.

ravjotbrar (changes requested):
Let's add some tests for the node-utilization logic. Also some derived logic can be moved to selectors.

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
```diff
@@ -46,7 +65,7 @@ export function ClusterNode({
   // Match the db-aware id scheme from buildConnectionId so the seed node
   // resolves to its status here, and node connects reuse that id
   // instead of creating a db id less duplicate.
-  const connectionId = buildConnectionId(primary.host, primary.port, clusterDb)
+  const connectionId = buildConnectionId(host, port, clusterDb)
```
ArgusLi, line 70: Why expand primary to host and port when primary is still being passed as an argument?

ArgusLi: Or rather, in what cases do host and port differ from primary? If it does, can we have more descriptive names?

apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -1,39 +1,56 @@
...
+import { getUtilizationLevel, type UtilizationLevel } from "./node-metrics"
+import type { PrimaryNode } from "@/state/valkey-features/cluster/clusterSlice"
 import { selectCluster } from "@/state/valkey-features/cluster/clusterSelectors"
 import { useAppDispatch } from "@/hooks/hooks"
-import { updateClusterData } from "@/state/valkey-features/cluster/clusterSlice"
-import { selectClusterAlias, selectClusterDb } from "@/state/valkey-features/connection/connectionSelectors"
+import { updateClusterData, stopClusterDataPolling } from "@/state/valkey-features/cluster/clusterSlice"
+import { selectClusterAlias } from "@/state/valkey-features/connection/connectionSelectors"
+
+type RoleFilter = "all" | "primary" | "replica"
```
ArgusLi: Can we create a Role Type mirroring Utilization Filter? This creates a single source of truth for roles and prevents future bugs.

apps/frontend/src/components/cluster-topology/Cluster.tsx
```diff
@@ -91,60 +151,94 @@ export function Cluster() {
...
+        </Select>
+        <Select
+          aria-label="Filter by utilization"
+          className="w-44 shrink-0"
+          icon={<ListFilter size={14} />}
+          onChange={(e) => setUtilizationFilter(e.target.value as UtilizationFilter)}
+          value={utilizationFilter}
+        >
+          <option value="all">All utilization</option>
+          <option value="low">Low</option>
```
ArgusLi, line 135: Should we replace these with the Utilization type values?

apps/frontend/src/components/cluster-topology/node-metrics.ts
```diff
@@ -0,0 +1,42 @@
+import { CPU_HIGH_THRESHOLD, CPU_NORMAL_THRESHOLD,
+  MEMORY_HIGH_THRESHOLD, MEMORY_NORMAL_THRESHOLD } from "@common/src/constants.ts"
+
+export type UtilizationLevel = "low" | "normal" | "high"
+
+const LEVEL_RANK: Record<UtilizationLevel, number> = { low: 0, normal: 1, high: 2 }
+
+// Memory and CPU get their own bands: a cache is meant to sit near its
+// memory limit, but a hot event loop is not meant to sit near one core.
+function levelFor(
```
ArgusLi: Should this be renamed `getLevelFor`?

apps/frontend/src/components/cluster-topology/node-metrics.ts
```diff
@@ -0,0 +1,42 @@
...
+  percent: number | null | undefined,
+  normalThreshold: number,
+  highThreshold: number,
+): UtilizationLevel | null {
+  if (percent === null || percent === undefined) return null
```
ArgusLi: We can replace this with `percent == null` - it catches both null and undefined cases.

apps/frontend/src/components/cluster-topology/node-metrics.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (percent >= highThreshold) return "high"
```
ArgusLi: Can we replace this with a switch on percent?

apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
```diff
@@ -26,11 +26,28 @@ export interface ParsedNodeInfo {
...
   used_cpu_sys: string | null;
   instantaneous_ops_per_sec: string | null;
   total_commands_processed: string | null;
   role: string | null;
   connected_clients: string | null;
+  keyspace_hits: string | null;
+  keyspace_misses: string | null;
+}
+
+export type MemoryBasis = "maxmemory" | "total_system_memory" | "none"
```
ArgusLi, line 52: maxmemory should be max_memory

apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
```diff
@@ -26,11 +26,28 @@ export interface ParsedNodeInfo {
   uptime_in_days: string | null;
   tcp_port: string | null;
   used_memory_human: string | null;
+  used_memory: string | null;
+  maxmemory: string | null;
```
ArgusLi, line 42: maxmemory should be max_memory

apps/server/src/node-utilization.ts
```diff
@@ -0,0 +1,216 @@
+import * as R from "ramda"
+import { type ParsedClusterInfo } from "./utils"
+
+export type MemoryBasis = "maxmemory" | "total_system_memory" | "none"
```
ArgusLi, line 4: maxmemory should be max_memory

apps/frontend/src/components/cluster-topology/Cluster.tsx
nassery318: Total nodes include both primary and replicas, however when it comes to flagged nodes we are looking into primaries only - client.info() routes to primaries only, so we have no replica metrics at all. So flagged / total would be misleading today.

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
nassery318: Removed both. 0 on 139 was actually wrong: memoryLimit > 0 made a missing limit show as ∞, which says the node has no limit set. They show - now

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
nassery318: addressed above

apps/frontend/src/components/cluster-topology/Cluster.tsx
nassery318: addressed

apps/frontend/src/components/cluster-topology/Cluster.tsx
nassery318: addressed

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
nassery318: They only differ on replica rows, where primary isn't used - it just carries connection settings (tls, username, authType) that only exist on PrimaryNode. Renamed it to primaryConfig.

apps/frontend/src/components/cluster-topology/Cluster.tsx
nassery318: added NodeRole and reused everywhere for consistency and single source of truth

apps/frontend/src/components/cluster-topology/node-metrics.ts
nassery318: renamed to getLevelFor so it is consistent.

apps/frontend/src/components/cluster-topology/node-metrics.ts
nassery318: switch matches exact values, but these are ranges ("above X"), so there's nothing to match on. It would have to be switch (true), which is harder to read for the same result.

apps/frontend/src/state/valkey-features/cluster/clusterSlice.ts
nassery318: maxmemory is Valkey's actual field name from the INFO command — it has no underscore, while total_system_memory does. same below comments

apps/server/src/connection.ts
nassery318: True, it was a redundant check - removed it. Moved the CPU sample cleanup inside the close block and it now clears all of the cluster's nodes, so a reconnect doesn't diff against stale CPU data

apps/frontend/src/components/cluster-topology/node-metrics.ts
nassery318: replaced

apps/server/src/connection.ts
```diff
@@ -749,9 +750,17 @@ export function teardownConnection(
       console.error(`Error closing connection ${connectionId}:`, error)
     }
 
-    if (connection.clusterId && !isWebMode) {
-      delete clusterNodesRegistry[connection.clusterId]
-      clusterCredentials.delete(connection.clusterId)
+    if (connection.clusterId) {
```
ravjotbrar, line 753: Does this mean we only clear cpu samples for clusters? What about standalone? I believe the fix is to move the first clearCpuSamples outside of this check.

nassery318: Yes, only cluster connections write to cpuSamples — the one set is at node-utilization.ts:194, reached only from setClusterDashboardData. Standalone connections gets its numbers from the metrics sidecar and never touches that map.

ravjotbrar: Ah right, brain fart on my end. It does seem like a redundant call. The second clear overlaps the first one, right?

nassery318: you are right, the first one was for the seed node (we manually type and connect to a cluster through that node) but since the second clearing case covers all nodes including the seed the first clear was redundant and removed.

ravjotbrar (approved):

PR #487, opened 2026-08-20
fixed unknown route blank page issue with a NOT Found page  and the e…

### LINKED ISSUE #443: Unknown route renders a blank white page instead of 404 or redirect
## Description

Navigating to an unmatched hash route (e.g. `#/this-route-does-not-exist`) renders a completely blank page with no nav rail, no content, no redirect, and no 404 message. The `#root` element has zero children.

## Reproduction

1. Navigate to `http://localhost:8080/#/nonexistent`
2. Observe a blank white page with no way to navigate back except editing the URL

## Impact

Mistyped URLs or stale bookmarks dead-end the user. Also affects the Electron `Cmd+4` shortcut which points to `/monitoring` (a route that no longer exists).

## Suggested Fix

Add a catch-all `*` route in `main.tsx` that renders a NotFound page or redirects to `/connect`. Add an error boundary as well.

nassery318:
…lectron shortcut

## Description

This PR adds a Not Found page for unknown routes and fixes keyboard shortcut navigation for Electron build. 

  - **Unknown routes rendered a blank page.** There was no catch-all route, so
    anything unmatched (e.g. `#/typo`) rendered nothing at all. Added a `NotFound`
    page inside the app layout, so it keeps the sidebar and shows the unmatched
    path with a link back to Connections.
- **Cmd/Ctrl+4 did nothing.** The menu sends the `activity` route key, but the
    renderer's shortcut map still had a stale `monitoring` entry pointing at
    `/:id/monitoring`, a route that doesn't exist. Mapped it to the real
    `/:clusterId/:id/activity` and `/:id/activity` routes, so it now works for both
    cluster and standalone connections.

Why a Not Found Page instead of the redirect to the /connect ?
A redirect hides the mistake - you just end up on Connections with no idea why. The 404 Not Found page tells you the address was wrong, so you can fix it.

### Change Visualization

image

### DIFF
--- a/apps/frontend/src/hooks/useShortcutNavigation.ts
+++ b/apps/frontend/src/hooks/useShortcutNavigation.ts
@@ -21,7 +21,7 @@ export const useShortcutNavigation = () => {
           sendcommand: clusterId ? `/${clusterId}/${id}/sendcommand` : `/${id}/sendcommand`,
           "cluster-topology": clusterId ? `/${clusterId}/${id}/cluster-topology` : "",
           settings: clusterId ? `/${clusterId}/${id}/settings` : `/${id}/settings`,
-          learnmore: `/${id}/learnmore`,
+          learnmore: clusterId ? `/${clusterId}/${id}/learnmore` : `/${id}/learnmore`,
         }
         : {
           connect: "/connect",
@@ -35,6 +35,6 @@ export const useShortcutNavigation = () => {
       }
     }
 
-    window.electronNavigation.onNavigate(handleNavigate)
+    return window.electronNavigation.onNavigate(handleNavigate)
   }, [navigate, id, clusterId, isConnected])
 }
--- a/apps/frontend/src/hooks/useShortcutNavigation.ts
+++ b/apps/frontend/src/hooks/useShortcutNavigation.ts
@@ -17,11 +17,11 @@ export const useShortcutNavigation = () => {
           connect: clusterId ? `/${clusterId}/${id}/connect` : `/${id}/connect`,
           dashboard: clusterId ? `/${clusterId}/${id}/dashboard` : `/${id}/dashboard`,
           browse: clusterId ? `/${clusterId}/${id}/browse` : `/${id}/browse`,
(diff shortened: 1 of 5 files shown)


apps/frontend/src/hooks/useShortcutNavigation.ts
```diff
@@ -17,11 +17,11 @@ export const useShortcutNavigation = () => {
           connect: clusterId ? `/${clusterId}/${id}/connect` : `/${id}/connect`,
           dashboard: clusterId ? `/${clusterId}/${id}/dashboard` : `/${id}/dashboard`,
           browse: clusterId ? `/${clusterId}/${id}/browse` : `/${id}/browse`,
-          monitoring: clusterId ? `/${clusterId}/${id}/monitoring` : `/${id}/monitoring`,
+          activity: clusterId ? `/${clusterId}/${id}/activity` : `/${id}/activity`,
           sendcommand: clusterId ? `/${clusterId}/${id}/sendcommand` : `/${id}/sendcommand`,
           "cluster-topology": clusterId ? `/${clusterId}/${id}/cluster-topology` : "",
           settings: clusterId ? `/${clusterId}/${id}/settings` : `/${id}/settings`,
-          learnmore: clusterId ? `/${clusterId}/${id}/learnmore` : `/${id}/learnmore`,
```
ravjotbrar, line 24: Why are we removing this?

nassery318: This is a common path, not exclusive to an instance or a cluster same as settings now. 

as seen from the main.tsx:
<Route element={<Settings />} path="/settings" />
<Route element={<LearnMore />} path="/learnmore" />

But adjusted it back as it keeps the user within the bounds of the cluster and standalone instances.

ravjotbrar: This isn't an issue from your PR, but there's a bug where clicking on learn more hides the cluster topology view from the sidebar. Can we address that in this PR as well, since we're already making changes there?

ArgusLi (dismissed):

ravjotbrar (approved):

PR #489, opened 2026-08-20
Validate connection port (1-65535) with inline error

### LINKED ISSUE #444: Port field accepts invalid values (0, -1, 99999, decimals)
## Description

The Port input in the Add Connection modal has no `min`, `max`, or `step` attributes. Values like `0`, `-1`, `99999`, and `6379.5` are all accepted and leave the Connect button enabled, resulting in opaque connection failures.

## Reproduction

1. Open Add Connection modal
2. Enter `-1` or `99999` or `6379.5` in the Port field
3. Observe the Connect button remains enabled
4. Clicking Connect produces an unhelpful connection error

## Suggested Fix

Add `min="1" max="65535" step="1"` to the port input. Validate integer range on submit with a field-level error message.

ravjotbrar:
Closes #444

## Problem
The Port field in the Add Connection modal had no bounds and no validation, so `0`, `-1`, `99999`, and `6379.5` were all accepted and left Connect enabled, producing opaque connection failures.

## Fix
- Added `min={1} max={65535} step={1}` to the port input.
- On submit, an invalid port shows an inline **"Invalid port"** message beneath the field (mirrors the existing `dbError` pattern) and blocks the connect dispatch.

### DIFF
--- a/apps/frontend/src/components/ui/connection-modal.tsx
+++ b/apps/frontend/src/components/ui/connection-modal.tsx
@@ -29,7 +29,6 @@ interface ConnectionModalProps {
   dbError?: string
   /** Informational hint under the Database field (e.g. the server's valid range). */
   dbHint?: string
-  portError?: string
 }
 
 export function ConnectionModal({
@@ -48,7 +47,6 @@ export function ConnectionModal({
   showVerifyTlsCertificate = true,
   dbError,
   dbHint,
-  portError,
 }: ConnectionModalProps) {
 
   return (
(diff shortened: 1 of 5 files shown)


apps/frontend/src/components/ui/connection-modal.tsx
```diff
@@ -147,14 +149,22 @@ export function ConnectionModal({
                       <Label className="block mb-1" htmlFor="port">Port</Label>
                       <Input
```
nassery318, line 148: I guess we should be using the ui/number-input.tsx as it is purpose build for min, max and steps.

ravjotbrar: Simplified it

ravjotbrar: Only reason I don't want to use number-input is that it clamps rather than surfacing an error

nassery318: better and simpler

nassery318 (approved):

PR #494, opened 2026-08-26
metric server binds to 127.0.0.1 for all modes except K8

nassery318:
## Description

Metrics processes bind on `0.0.0.0` with no auth on any endpoint, so anyone on the LAN can hit `/monitor`, `/commandlog` and `POST /update-config` directly. #466 fixed this on the main server but not on the metrics children.

The server now spawns them with `METRICS_BIND_HOST=127.0.0.1`. It already reaches them over
loopback (they advertise `127.0.0.1` by default), so nothing changes functionally.

### Change Visualization

### DIFF
--- a/docs-site/src/content/docs/configuration/metrics.md
+++ b/docs-site/src/content/docs/configuration/metrics.md
@@ -134,9 +134,11 @@ Port of the Valkey Admin server.
 
 ### `METRICS_BIND_HOST`
 
-Network interface the metrics HTTP server binds to. In a container you almost always want `0.0.0.0`; on a developer machine you might prefer `127.0.0.1`.
+Network interface the metrics HTTP server binds to. The metrics endpoints have no auth, so keep them on loopback unless the server must reach them from another host — which is the Kubernetes sidecar case.
 
-- **Default:** `0.0.0.0`
+- **Default:** `127.0.0.1`; `0.0.0.0` for the Kubernetes sidecar
+
+If you set a routable `METRICS_ADVERTISE_HOST`, set this to match.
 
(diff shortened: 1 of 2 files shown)


docs-site/src/content/docs/configuration/metrics.md
```diff
@@ -134,9 +134,11 @@ Port of the Valkey Admin server.
 
 ### `METRICS_BIND_HOST`
 
-Network interface the metrics HTTP server binds to. In a container you almost always want `0.0.0.0`; on a developer machine you might prefer `127.0.0.1`.
+Network interface the metrics HTTP server binds to. The metrics endpoints have no auth, so keep them on loopback unless the server must reach them from another host — which is the Kubernetes sidecar case.
 
-- **Default:** `0.0.0.0`
+- **Default:** `127.0.0.1`; `0.0.0.0` for the Kubernetes sidecar
+
+If you set a routable `METRICS_ADVERTISE_HOST`, set this to match it.
```
ravjotbrar: We could probably remove this, since it's not entirely accurate.

nassery318: I believe the default part holds as we are now changing the default for metrics server from open to all in a network to internal (127.0.0.1) and the K8 part due to inner pod connections.

The second part may be removed.

ravjotbrar: Referring to the second part

nassery318: removed

ravjotbrar (approved):

PR #500, opened 2026-08-27
Scope Valkey client reuse to the owning session

ravjotbrar:
## Summary

Connection reuse (`getExistingConnection` / `getExistingClusterClient`) matched a cached Valkey client purely by resolved `host:port:db` (or cluster membership), so a session could be handed a client that a *different* session established, just by connecting to the same endpoint. This scopes client reuse — and connection retry — to the session that owns the connection.

## Details

- `getExistingConnection` — only returns a cached client when the session owns the matched connection id (`isConnectionAuthorized`).
- `getExistingClusterClient` — only reuses the shared cluster client when the session owns the matched sibling node. This preserves connecting to another node of a cluster you already connected to (owning any node authorizes reuse of the shared client), while a session that never connected to the cluster gets its own fresh client.
- **Retry** (`isRetry`) is now gated on ownership too, mirroring the existing `isResume` check. Retry reconnects an existing connection and, for clusters, closes + repoints the shared client — so one session's retry must not be able to close or repoint a connection another session owns (e.g. a same-endpoint collision on a shared instance). The frontend only retries connections it previously established, so no legitimate retry is rejected.
- `sessionId` is now a required part of the connect path (threaded through `connectToValkey` → `connectToValkeyLocked` → the reuse helpers); the connect handler fails closed if it is somehow absent.

Same-session, same-cluster node browsing is unchanged: the session owns its nodes after connecting, so it keeps reusing its own cluster client (no extra re-authentication per node).

## Testing

- `npm test --workspace=server` — 205 pass
- `npm run build` — server + frontend
- `npx eslint .` — clean

Regression tests: `getExistingConnection` does not return a client the calling session is not authorized for; `getExistingClusterClient` reuses for an owned sibling node and denies an unowned one; existing reuse/dedup tests updated to set up an authorized session.

### DIFF
--- a/apps/server/src/connection.ts
+++ b/apps/server/src/connection.ts
@@ -690,11 +690,8 @@ export async function getExistingConnection(
   // could inherit another session's authenticated client by supplying a matching
   // endpoint with the wrong/blank credentials (credential-blind reuse). The matched id
   // is the exact node being (re)connected to; owning it is required to reuse its client.
-  const isMatchAuthorized = (matchedId: string): boolean =>
-    isConnectionAuthorized(sessionId, matchedId)
-
   if (clients.has(connectionId)) {
-    return isMatchAuthorized(connectionId) ? clients.get(connectionId) : undefined
+    return isConnectionAuthorized(sessionId, connectionId) ? clients.get(connectionId) : undefined
   }
 
   const db = connectionDetails.db
@@ -703,7 +700,7 @@ export async function getExistingConnection(
     .find((key) => clients.has(key))
 
   if (!existingConnectionId) return undefined
-  return isMatchAuthorized(existingConnectionId) ? clients.get(existingConnectionId) : undefined
+  return isConnectionAuthorized(sessionId, existingConnectionId) ? clients.get(existingConnectionId) : undefined
 }
 
 export async function closeMetricsServer(
--- a/apps/server/src/connection.ts
+++ b/apps/server/src/connection.ts
@@ -26,6 +26,7 @@ import {
 import { subscribe } from "./node-watchers"
 import { clearCpuSamples } from "./node-utilization"
 import { createClusterValkeyClient, createStandaloneValkeyClient } from "./valkey-client"
+import { isConnectionAuthorized } from "./session"
 
 export type ConnectionContext = {
   clients: Map<string, { client: GlideClient | GlideClusterClient; clusterId?: string }>
@@ -138,6 +139,7 @@ export async function connectToValkey(
     connectionDetails: ConnectionDetails
     connectionId: string
     isRetry?: boolean
+    sessionId: string
   },
 ) {
   const { connectionId } = payload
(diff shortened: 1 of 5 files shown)


apps/server/src/connection.ts
```diff
@@ -668,29 +671,39 @@ function sendClusterConnectFulfilled(ws: WebSocket, payload: ClusterConnectFulfi
...
+  // Only reuse a cached client this session is authorized for. Otherwise a session
+  // could inherit another session's authenticated client by supplying a matching
+  // endpoint with the wrong/blank credentials (credential-blind reuse). The matched id
+  // is the exact node being (re)connected to; owning it is required to reuse its client.
+  const ownsMatch = (matchedId: string): boolean =>
+    isConnectionAuthorized(sessionId, matchedId)
+
+  if (clients.has(connectionId)) {
+    return ownsMatch(connectionId) ? clients.get(connectionId) : undefined
+  }
```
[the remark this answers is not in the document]
ravjotbrar: We may need to scope clients per session rather than a global map to avoid this. Not in this PRs scope though.

apps/server/src/connection.ts
```diff
@@ -668,29 +671,39 @@ function sendClusterConnectFulfilled(ws: WebSocket, payload: ClusterConnectFulfi
...
-  if (clients.has(connectionId)) return clients.get(connectionId)
+  // Only reuse a cached client this session is authorized for. Otherwise a session
+  // could inherit another session's authenticated client by supplying a matching
+  // endpoint with the wrong/blank credentials (credential-blind reuse). The matched id
+  // is the exact node being (re)connected to; owning it is required to reuse its client.
+  const isMatchAuthorized = (matchedId: string): boolean =>
+    isConnectionAuthorized(sessionId, matchedId)
+
+  if (clients.has(connectionId)) {
+    return isMatchAuthorized(connectionId) ? clients.get(connectionId) : undefined
```
ArgusLi: Can we unfold this? `isMatchAuthorized` is essentially a rewording of `isConnectionAuthorized` with 1 less variable. This creates a layer of complexity for very little tradeoff.

apps/server/src/connection.ts
```diff
@@ -668,29 +671,39 @@ function sendClusterConnectFulfilled(ws: WebSocket, payload: ClusterConnectFulfi
...
+  }
 
   const db = connectionDetails.db
   const existingConnectionId = resolvedAddresses
     .map((address) => buildConnectionId(address, connectionDetails.port, db))
     .find((key) => clients.has(key))
 
-  return existingConnectionId ? clients.get(existingConnectionId) : undefined
+  if (!existingConnectionId) return undefined
+  return isMatchAuthorized(existingConnectionId) ? clients.get(existingConnectionId) : undefined
```
ArgusLi: Same as above. Alternatively, you can fold the entire line into it's own function as it's reused and give a better explanation.

apps/server/src/connection.ts
ravjotbrar: Good point. Unfolded.

ArgusLi (approved):

PR #501, opened 2026-08-28
Refactor config to harden update-config

ArgusLi:
## Description

Refactors config to consolidate and harden.
- Unify epic identity on name
    - Collapse name, type, file_prefix into a single name
    - Remove file_prefix from update-config to harden
- Harden update-config 
    - Rework validator into zod validator
    - Update shape from `epic: { name, ... }` to `epics: { <name>: { ... } }`
    - Updates are all-or-nothing for named epics
- Isolate config updates for each process
    - Previously, metrics children were spawned with the same `CONFIG_PATH` which led to them each overwriting the other's config files.
    - This change stages them config changes through `config.yml.<pid>.tmp`.
-  Patch config updates as a YAML document rather than re-serializing runtime config
- Apply patch in memory before persisting, so deployments with read-only memory can still use the changes, without persistence.

### DIFF
--- a/apps/metrics/src/config-schema.js
+++ b/apps/metrics/src/config-schema.js
@@ -1,0 +1,71 @@
+import { z } from "zod"
+import { EPIC_FIELD_BOUNDS } from "../../../common/src/constants.js"
+
+/**
+ * Request schema for POST /update-config.
+ *
+ * The endpoint tunes collection settings on epics that already exist in
+ * `config.yml`; it is not a general config writer. Everything is default-deny:
+ *
+ *  - the only accepted top-level key is `epics`
+ *  - `epics` is a map keyed by epic name, so no entry can be created, removed
+ *    or renamed — the key is used purely to look up an existing epic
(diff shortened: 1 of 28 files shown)


ravjotbrar (approved):

apps/metrics/src/config-schema.js
```diff
@@ -0,0 +1,71 @@
...
+ *    what keeps API input out of every filesystem path (see ndjson-writer.js)
+ *  - unknown fields are rejected rather than ignored, so a typo fails loudly
+ *    instead of silently doing nothing
+ *
+ * Field bounds live in `common/src/constants.ts` so the UI inputs that produce
+ * these values and this validator cannot drift apart.
+ */
+
+// Epic names double as NDJSON filename prefixes, so keep them to a safe token.
+const EPIC_NAME_PATTERN = /^[a-z0-9_]+$/
```
nassery318, line 23: maybe we can export this so we don't declare it again in config.js line 30. single source of truth.

PR #503, opened 2026-08-31
implementation of cluster hot slots version 1

nassery318:
## Description

Adds a Slot Heatmap view to the Hot Keys feature and threads slotId through the full pipeline (metrics collector → server aggregation → Redux → UI). The existing single-view "Node Heatmap" modal becomes a tabbed Heatmap modal with Nodes and Slots tabs.

  - Hot-slots scanner now keeps each key's slotId; returns [key, freq, slotId] instead of [key, freq].
  - enrichHotKeys passes trailing tuple fields through, so slotId survives TTL/size enrichment.
  - Server cluster aggregation carries slotId as a 6th tuple element (next to nodeId), keeping first-seen slot per key.
  - HotKeyEntry type and hotKeysSlice state widened to [key, count, size, ttl, nodeId?, slotId?].
  - Heatmap modal converted to tabbed Nodes / Slots view; enlarged to 90vw × 80vh; per-tab title/description.
  - Slots tab only shows in hot-slots mode when at least one entry has a slotId.
  - Node-heatmap logic extracted unchanged into new node-heatmap.tsx.
  - New shared heatmap-scale.ts (color/bucket/ratio helpers) and heatmap-legend.tsx (clickable legend).
  - New slot-heatmap.tsx: groups keys by slot, slot-tile grid + selected-slot key list (size/TTL/owner); key click closes modal and navigates.
  - SlotHeatmap shows empty state and "Partial — N nodes failed to report" from failedNodeCount (nodeErrors.length).
  - hot-keys.tsx passes failedNodeCount, onKeyClick, and showSlots into the modal.
  - Toolbar button relabeled "Node Heatmap" → "Heatmap".

### Change Visualization

Screenshot 2026-08-31 at 4 36 21 PM

### DIFF
--- a/apps/frontend/src/components/activity-view/hotkeys/heatmap-legend.tsx
+++ b/apps/frontend/src/components/activity-view/hotkeys/heatmap-legend.tsx
@@ -1,0 +1,40 @@
+import { Typography } from "../../ui/typography"
+import { getColor, LEGEND_STEPS } from "./heatmap-scale"
+
+interface HeatmapLegendProps {
+  label: string
+  selectedBuckets: Set<number>
+  onToggle: (step: number) => void
+}
+
+export function HeatmapLegend({ label, selectedBuckets, onToggle }: HeatmapLegendProps) {
+  const hasBucketFilter = selectedBuckets.size > 0
+
+  return (
+    <div className="flex items-center justify-between gap-4">
+      <Typography variant="bodyXs">{label}</Typography>
+      <div className="flex gap-1 shrink-0">
+        {LEGEND_STEPS.map((step) => {
+          const isSelected = selectedBuckets.has(step)
+          return (
+            <button
+              aria-label={`Filter band ${LEGEND_STEPS.indexOf(step) + 1}`}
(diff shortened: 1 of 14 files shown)


ravjotbrar (approved):

ravjotbrar (changes requested):
group.nodeId is a string (typed at line 17, set from row[4] at line 46). flatMap spreads its callback's return value — and strings are iterable,
  so returning a string like "127-0-0-1-7000" flattens into individual characters ["1","2","7","-",...].

nassery318:
> group.nodeId is a string (typed at line 17, set from row[4] at line 46). flatMap spreads its callback's return value — and strings are iterable, so returning a string like "127-0-0-1-7000" flattens into individual characters ["1","2","7","-",...].

flatMap only unpacks arrays, not strings - so this was already returning whole node ids, not characters. Switched to .map().filter() anyways since it reads more clearly.

Console log results:
Nodes ::: (2) ['192-168-18-6-7002', '192-168-18-6-7003']

PR #506, opened 2026-09-03
set clientInfoTag and upgrade glide to version 2.5.2

### LINKED ISSUE #505: Upgrade to valkey-glide 2.5.2 and set clientInfoTag for adoption tracking
## Summary

Upgrade `@valkey/valkey-glide` to 2.5.2 and use its new `clientInfoTag` option so Valkey Admin's connections are identifiable on the server, and distinguishable by which process opened them and which deployment mode that process is running in.

`clientInfoTag` appends a parenthesized tag to the library name a client reports — e.g. `GlideJS(valkey-admin-web:1.2.0)` — while preserving GLIDE's own identity. Don't override the library name itself; that replaces GLIDE's identity rather than tagging alongside it.

| Process | Deployment mode | Tag |
|---|---|---|
| server | Electron | `valkey-admin-electron:<version>` |
| server | Web | `valkey-admin-web:<version>` |
| server | K8s | `valkey-admin-k8s:<version>` |
| metrics | Electron | `valkey-admin-metrics-electron:<version>` |
| metrics | Web | `valkey-admin-metrics-web:<version>` |
| metrics | K8s | `valkey-admin-metrics-k8s:<version>` |

The metrics collector needs the mode split too — it is not a single deployable. It runs as a child of the server in Electron and Web, and as a sidecar in K8s. With one flat tag, a desktop user's collectors are indistinguishable from a K8s fleet's.

`<version>` is the single app version already used for releases. The individual app packages carry no version of their own, so there is no Electron-vs-backend drift to resolve.

### 1. Dependency upgrade

Land the upgrade separately from the tagging change so any regression is attributable.

2.4 → 2.5 is not a routine minor: it changed connection handling, TLS, and cluster recovery — all paths this app exercises. Confirm connection, health-check, and cluster behavior still hold on standalone and cluster, and that the client still works in every artifact the project ships, on each supported architecture.

### 2. Tagging

- Every connection the app opens carries the tag for its process and deployment mode.
- An unset or unrecognized deployment mode falls back to the Web tag.
- Derive the tag from the deployment mode the process already knows; no new detection mechanism.
- Existing client names are unchanged — this must not affect `CLIENT SETNAME` behavior.
- A malformed tag fails loudly at startup, rather than silently or once per connection attempt.
- The version in the tag comes from the single app version rather than a hardcoded string, and is correct in dev, in the containers, and in the packaged desktop app.
- Shipped artifacts must not pick up build-time metadata — dependency lists, scripts, packaging or signing configuration — as a side effect of reading that version.

### 3. Deployment coverage

The tag must be correct in every shape the project ships: desktop app, container/web service, and the K8s app plus its metrics sidecars. Where a deployment doesn't yet hand the deployment mode to a process that now needs it, supply it.

### 4. Docs

Update user-facing docs in the same change: anything stating the GLIDE client version, release notes, third-party notices, and a note on the client identity operators will now see for Valkey Admin connections on their own servers.

## Tests

- Each deployment mode produces its expected tag, for both processes and every client they open; unset and unrecognized modes fall back.
- GLIDE's default reported identity is preserved, and existing client names are unchanged.
- Connections from different deployment modes are distinguishable from one another on a server.
- Existing connection and health-check behavior is unaffected by the upgrade, standalone and cluster.
- Verified against a live server in each deployment shape. K8s sidecar registration currently fails and the sidecar exits after repeated attempts; the tag is observable before it exits, so don't block on that path.

## Refs

`clientInfoTag` shipped in [glide 2.5.2](https://github.com/valkey-io/valkey-glide/releases/tag/v2.5.2) via backport [#6901](https://github.com/valkey-io/valkey-glide/pull/6901); originally [#6389](https://github.com/valkey-io/valkey-glide/pull/6389), extended in [#6755](https://github.com/valkey-io/valkey-glide/pull/6755). Upstream docs pending: [valkey-glide-docs#321](https://github.com/valkey-io/valkey-glide-docs/pull/321).

nassery318:
## Description

Upgrade valkey-glide to 2.5.2 and set clientInfoTag

  Closes #505

  Bumps @valkey/valkey-glide from ^2.4.0 to ^2.5.2 in apps/server and apps/metrics and sets the new clientInfoTag option on every Glide client so Valkey Admin connections are
  identifiable in CLIENT LIST / CLIENT INFO.

  Tags:
  - server, derived from DEPLOYMENT_MODE: valkey-admin-electron:<version>, valkey-admin-k8s:<version>, valkey-admin-web:<version> (default when unset)
  - metrics: valkey-admin-metrics:<version>

  Resulting lib-name looks like GlideJS(valkey-admin-web:1.2.0).

  Notes:
  - The version comes from the root package.json rather than per-app manifests, since all apps share one release number.
  - The existing env var is DEPLOYMENT_MODE (compared against the DEPLOYMENT_TYPE constants in valkey-common), so that is what drives the suffix.
  - clientName values are unchanged and libName is not set, so Glide's default identity is preserved.
  - Updated the metrics client test to expect the new option.

  Verified with the server and metrics unit suites and CLIENT LIST against a local Valkey.

### Change Visualization

Electron:
Screenshot 2026-09-03 at 12 45 57 PM

WEB:
image


apps/metrics/src/valkey-client.js
```diff
@@ -1,4 +1,7 @@
 import { GlideClient, GlideClusterClient, ServiceType, NodeDiscoveryMode } from "@valkey/valkey-glide"
+import pkg from "../../../package.json" with { type: "json" }
+
+const clientInfoTag = `valkey-admin-metrics:${pkg.version}`
```
alexey-temnikov: `valkey-admin-metrics:<version>` is identical in all three deployment modes. Verified against one server: a Web-mode app's spawned collector and a K8-mode sidecar both report `GlideJS(valkey-admin-metrics:1.2.0)` — indistinguishable.

The metrics child already inherits `DEPLOYMENT_MODE` via the `...process.env` spread in `metrics-orchestrator.ts:479`, so reusing the same suffix helper yields `valkey-admin-metrics-{electron,web,k8s}` at no cost. K8s sidecars would additionally need `DEPLOYMENT_MODE=K8` in `examples/k8s/valkey-statefulset.yaml` and `valkey-statefulset-sidecar-patch.yaml` — neither sets it today.

apps/server/src/valkey-client.ts
```diff
@@ -1,4 +1,6 @@
 import { GlideClient, GlideClusterClient, NodeDiscoveryMode, type ServerCredentials } from "@valkey/valkey-glide"
+import { DEPLOYMENT_TYPE } from "valkey-common"
+import pkg from "../../../package.json" with { type: "json" }
```
alexey-temnikov: Importing the whole manifest makes esbuild inline **all** of root `package.json` into `apps/server/dist/index.js` — the bundle `Dockerfile.app` runs. The runtime image then ships devDependencies, every npm script, and the full electron-builder `build` block (signing hooks `mac_build/notarize.js`, `linux_build/sign.js`, entitlements paths).

Vite's output tree-shakes to `const pkg = { version: "1.2.0" }`, so only the tsup path is affected. A build-time `define` (as `apps/frontend/vite.config.ts:16` already does) or importing just the version avoids it.

alexey-temnikov (changes requested):

apps/metrics/src/valkey-client.js
nassery318: metrics tag is mode-aware via a helper shared from valkey-common, and the k8s example sidecars set DEPLOYMENT_MODE=K8.

apps/server/src/valkey-client.ts
nassery318: server now imports only the version field, tsup output verified free of manifest content.

apps/metrics/src/valkey-client.js
```diff
@@ -1,4 +1,8 @@
 import { GlideClient, GlideClusterClient, ServiceType, NodeDiscoveryMode } from "@valkey/valkey-glide"
+import { deploymentSuffix } from "valkey-common"
+import { version } from "../../../package.json"
```
alexey-temnikov: Importing `package.json` directly via relative paths still looks anti-pattern.

apps/metrics/package.json
```diff
@@ -15,7 +15,7 @@
     "@aws-sdk/util-format-url": "^3.972.9",
     "@smithy/protocol-http": "^5.3.13",
     "@smithy/signature-v4": "^5.3.13",
-    "@valkey/valkey-glide": "^2.4.0",
+    "@valkey/valkey-glide": "^2.5.2",
```
alexey-temnikov, line 18: Please update the user-facing docs with this upgrade. `docs-site/src/content/docs/development/platform-support.md` still states GLIDE 2.4, and operators should be told what `CLIENT LIST` / `CLIENT INFO` identity Valkey Admin now reports.

alexey-temnikov (changes requested):
The current head still needs changes: version loading must work in the native metrics runtime as well as every bundled artifact, with regression coverage and the required documentation updates.

apps/metrics/src/valkey-client.js
alexey-temnikov: Importing `package.json` directly via relative paths still looks anti-pattern.
 
Could the version come from a shared build/runtime contract covering Vite, tsup, and tsx?

nassery318: 1. **New `common/src/version.ts`** — a one-line `APP_VERSION` constant, written by `scripts/sync-version.mjs` from the root `package.json`. Runs as part of `common`'s build,
  so the root manifest stays the only place you bump a version.
  2. **Exported it** from `common/src/index.ts`.
  3. **Server and metrics now read the version from `valkey-common`** — the same import they already used for `deploymentSuffix`, so this rides on the existing contract rather
  than adding a second one.
  4. **Dropped all three `../../../package.json` imports** — server client, metrics client, and the metrics test.

common/scripts/sync-version.mjs
```diff
@@ -0,0 +1,7 @@
+import { readFileSync, writeFileSync } from "node:fs"
+import { fileURLToPath } from "node:url"
+import { dirname, resolve } from "node:path"
+
+const root = resolve(dirname(fileURLToPath(import.meta.url)), "..")
+const { version } = JSON.parse(readFileSync(resolve(root, "package.json"), "utf-8"))
+writeFileSync(resolve(root, "common/src/version.ts"), `export const APP_VERSION = "${version}"\n`)
```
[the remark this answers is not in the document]
nassery318: Already fixed, though with a different approach than suggested.

  Rather than adding `COPY scripts ./scripts` to each builder stage, the generator moved from `scripts/sync-version.mjs` to `common/scripts/sync-version.mjs`. The existing `COPY
  common ./common` in all three builders already carries it, so no Dockerfile change is needed.

  - `common/package.json` calls `node scripts/sync-version.mjs` (npm sets cwd to `common/`)
  - the script resolves the repo root as `../..` from its own location, so it's cwd-independent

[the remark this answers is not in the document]
nassery318: The script already uses `../..` on this branch — the `..` in your comment isn't what's pushed.

  `common/scripts/sync-version.mjs` line 5:

  ```js
  const root = resolve(dirname(fileURLToPath(import.meta.url)), "../..")

  Confirmed against the pushed commit via git show HEAD:common/scripts/sync-version.mjs, not just the working tree.

  Actual behavior when run:
  - resolves to the repository root
  - reads the root package.json
  - writes common/src/version.ts

  No common/common/ path is produced, and none exists in the tree. npm run build:all with the root scripts/ directory removed (simulating the Docker context) exits 0, and the
  integration build is green.

  Looks like this was generated against an earlier revision of the file.

common/scripts/sync-version.mjs
```diff
@@ -0,0 +1,7 @@
+import { readFileSync, writeFileSync } from "node:fs"
+import { fileURLToPath } from "node:url"
+import { dirname, resolve } from "node:path"
+
+const root = resolve(dirname(fileURLToPath(import.meta.url)), "../..")
+const { version } = JSON.parse(readFileSync(resolve(root, "package.json"), "utf-8"))
+writeFileSync(resolve(root, "common/src/version.ts"), `export const APP_VERSION = "${version}"\n`)
```
alexey-temnikov, line 7: nit:
```suggestion
writeFileSync(resolve(root, "common/src/version.ts"), `export const APP_VERSION = "${version}"\n`)

```

nassery318: I think the suggested change is identical to the current implementation. Is there a specific formatting change intended here?

alexey-temnikov:
One K8 artifact issue remains in `docker/Dockerfile.metrics`. Its builder currently runs `build:metrics` without first producing `common/dist`, so a clean checkout fails with:

```text
Failed to resolve entry for package "valkey-common"
```

The same command fails on the base revision, so this is a pre-existing deployment-path issue rather than a regression introduced here. It still blocks #505's requirement that the tag work in the K8 metrics sidecar artifact.

Suggested edit:

```diff
 RUN npm ci
+RUN npm run build:common
 RUN npm run build:metrics
```

I verified the added `build:common` step followed by `build:metrics` succeeds on the current PR head.

docs-site/src/content/docs/development/platform-support.md
```diff
@@ -7,7 +7,7 @@ Valkey Admin is designed to work across multiple platforms with varying levels o
 
 ## Valkey Version Compatibility
 
-Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) 2.4 as its client library and works with all versions of Valkey. Some features require newer versions:
+Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) as its client library and works with all versions of Valkey. Some features require newer versions:
```
alexey-temnikov:
```suggestion
Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) as its client library and works with all versions of Valkey. Some features require newer Valkey versions:
```

nassery318: adjusted

docs-site/src/content/docs/development/platform-support.md
```diff
@@ -7,7 +7,7 @@ Valkey Admin is designed to work across multiple platforms with varying levels o
 
 ## Valkey Version Compatibility
 
-Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) 2.4 as its client library and works with all versions of Valkey. Some features require newer versions:
+Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) as its client library and works with all versions of Valkey. Some features require newer Valkey versions:
```
alexey-temnikov, line 10: Removing the 2.4 pin fixes the stale version, but #505 also requires the operator-visible client identity, release notes, and third-party notices. `NOTICES` still lists GLIDE 2.3.1. This covers the identity text; please also add the release note and update `NOTICES` to 2.5.2.

```suggestion
Valkey Admin uses [Valkey GLIDE](https://github.com/valkey-io/valkey-glide) as its client library and works with all versions of Valkey. Some features require newer Valkey versions:

When inspecting `CLIENT LIST` or `CLIENT INFO`, operators can identify Valkey Admin connections by the reported `lib-name`, for example `GlideJS(valkey-admin-web:1.2.0)` for the server or `GlideJS(valkey-admin-metrics-k8s:1.2.0)` for a Kubernetes metrics collector. The deployment suffix is `electron`, `web`, or `k8s`.
```

alexey-temnikov (commented):
One clean-checkout startup fix remains. At the latest head, `npm ci && npm run dev` fails during `build:metrics` because `valkey-common/dist` has not been built. Running `npm run build:common` first makes the metrics build pass.

Please insert `npm run build:common &&` immediately before `npm run build:metrics` in the root `package.json` `dev` script. This failure is pre-existing, but #505 requires the generated application version to work in development.

alexey-temnikov (dismissed):

ravjotbrar (dismissed):

common/scripts/sync-version.mjs
ArgusLi: The suggested change is the new line at the end of the file. Can you add this?

ArgusLi (approved):

Issue #507, opened 2026-09-03
Support GCP Memorystore for Valkey with IAM authentication and in-transit encryption (private CA)


rezakaramad:
### Before submitting

- [x] I searched existing issues and didn't find a duplicate

## Summary

Valkey Admin cannot currently connect to a **GCP Memorystore for Valkey** instance that has [**IAM authentication**](https://docs.cloud.google.com/memorystore/docs/valkey/about-iam-auth) and [**in-transit encryption (TLS)**](https://docs.cloud.google.com/memorystore/docs/valkey/about-in-transit-encryption) enabled. Two things are missing:

1. **IAM authentication**: Memorystore uses short-lived **OAuth2 access tokens** as the Valkey `AUTH` password. Valkey Admin has an AWS ElastiCache IAM path but no GCP equivalent, and short-lived tokens must be **rotated** on live connections.
2. **Private CA verification** — Memorystore's TLS certificate is signed by a **Google-managed CA that is not in the system trust store**, and Valkey Admin has no way to supply a custom CA, so verified TLS is impossible today (you can only connect with verification disabled).

This issue documents how Memorystore for Valkey works and proposes a design to support it.

---

### IAM authentication

Memorystore integrates IAM with the Valkey `AUTH` command. Instead of a static password, the client presents a short-lived Google **OAuth2 access token**:

- **Credential** = an OAuth2 access token minted for the caller's Google identity (user or service account), scoped to `https://www.googleapis.com/auth/cloud-platform`.
- **Username** = **`default`** - this is the **only supported username**. Authenticating with any other username returns `-WRONGPASS invalid username-password pair`. (Google's own client sample returns `("default", token)`.) ([See here](https://docs.cloud.google.com/memorystore/docs/valkey/manage-iam-auth#error-messages))
- **Authorization** = the principal must hold the **`roles/memorystore.dbConnectionUser`** role (grants `memorystore.instances.connect`).
- **Token lifetime** = **~1 hour** by default (extendable up to 12 hours when generated explicitly). The token is verified **at connection time**:
  - An **already-authenticated connection keeps working** even after the token expires.
  - A **new or reconnecting connection requires a currently-valid token.**
- Google explicitly recommends **fetching tokens on demand** via a client credential provider / callback, and **not** hardcoding or over-caching them.

[...]

PR #508, opened 2026-09-03
feat: add GCP Memorystore for Valkey IAM auth and private-CA TLS support - ISSUE-507

rezakaramad:
## What

Adds support for **GCP Memorystore for Valkey** with **IAM authentication** and **in-transit encryption (private CA)**.

- **`gcp-iam` auth type** (server, spawned metrics collectors, frontend). Mints short-lived OAuth2 tokens from Application Default Credentials (Workload Identity / GCE metadata / `GOOGLE_APPLICATION_CREDENTIALS`), scoped to `cloud-platform`, and uses them as the `AUTH` password with the fixed `default` user (the only username Memorystore accepts).
- **Token rotation** via GLIDE `updateConnectionPassword(token, true)` before expiry — re-auths every cluster-node connection. GLIDE's native `iamConfig` is AWS-only, so GCP tokens are minted/refreshed by the app.
- **Private-CA TLS:** `VALKEY_CA_CERT_PATH` seeds the existing `caCertPath` field and is passed to GLIDE as `rootCertificates` (and to the iovalkey MONITOR stream as `tls.ca`). Needed because GLIDE's Rust-core TLS ignores Node's trust store / `NODE_EXTRA_CA_CERTS`.

Also: bumps `google-auth-library` to `^11`, adds server tests for the credential and CA wiring, and documents the new config.

## Design

Follows the approved design proposal in #507

## Notes for reviewers (AI-assisted)

Some of this was AI-assisted; flagging assumptions to sanity-check:
- `VALKEY_CA_CERT_PATH` seeds the existing `caCertPath` field (mirroring `VALKEY_TLS` → `tls`) rather than introducing a parallel concept; there's no per-connection CA **UI** yet (existing `caCertPath` TODO).
- Token refresh uses `setInterval`, consistent with existing backend periodic work (WS heartbeat, session cleanup, metrics ping); the epics rule applies to the frontend Redux store, which the server/metrics don't have.
- `default` is hardcoded as the IAM username per GCP docs (any other value returns `-WRONGPASS`).

## Testing

- Server unit tests (gcp-iam credential + CA `rootCertificates`) pass; lint/typecheck clean.
- Manually verified against a Memorystore for Valkey cluster (IAM + TLS) on GKE.

### DIFF
--- a/apps/metrics/package.json
+++ b/apps/metrics/package.json
@@ -16,6 +16,7 @@
     "@smithy/signature-v4": "^5.3.13",
     "@valkey/valkey-glide": "^2.4.0",
     "express": "^4.21.2",
+    "google-auth-library": "^11.0.2",
     "heap-js": "^2.7.1",
     "iovalkey": "^0.3.3",
     "ramda": "^0.31.3",
--- a/apps/metrics/src/index.js
+++ b/apps/metrics/src/index.js
@@ -15,6 +15,7 @@ import { bigKeysQuerySchema, cpuQuerySchema, memoryQuerySchema, parseQuery } fro
 import { sanitizeUrl } from "./utils/helpers.js"
 import { setupNdjsonCleaner, stopNdjsonCleaner } from "./effects/ndjson-cleaner.js"
 import { createValkeyClient } from "./valkey-client.js"
+import { GcpIAMProvider } from "./utils/gcp-iam-provider.js"
 import { scanBigKeys } from "./analyzers/scan-big-keys.js"
 
 async function main() {
@@ -30,6 +31,19 @@ async function main() {
 
   const client = await createValkeyClient(cfg)
   const ownNodeId = sanitizeUrl(`${process.env.VALKEY_HOST}-${process.env.VALKEY_PORT}`)
+
+  // GCP OAuth2 tokens expire ~1h; rotate the connection password before then so
+  // reconnects keep authenticating. AWS IAM refreshes natively inside Glide.
+  const gcpTokenRefresh = process.env.VALKEY_AUTH_TYPE === "gcp-iam"
+    ? setInterval(async () => {
+      try {
+        await client.updateConnectionPassword(await new GcpIAMProvider().getCredentials(), true)
+      } catch (err) {
+        console.error("[gcp-iam] token refresh error:", err.message)
+      }
+    }, 45 * 60 * 1000)
+    : undefined
+  gcpTokenRefresh?.unref?.()
 
   await setupNdjsonCleaner(cfg)
   await setupCollectors(client, cfg)
@@ -225,6 +239,7 @@ async function main() {
     try {
       await stopNdjsonCleaner()
       await stopCollectors()
+      if (gcpTokenRefresh) clearInterval(gcpTokenRefresh)
       if (client) {
         client.close()
       }
--- a/apps/metrics/src/valkey-client.js
+++ b/apps/metrics/src/valkey-client.js
@@ -1,4 +1,6 @@
 import { GlideClient, GlideClusterClient, ServiceType, NodeDiscoveryMode } from "@valkey/valkey-glide"
+import { readFileSync } from "node:fs"
+import { GcpIAMProvider } from "./utils/gcp-iam-provider.js"
 
 const SUPPORTED_VALKEY_MODES = new Set(["standalone", "cluster"])
 
@@ -30,22 +32,35 @@ export const createValkeyClient = async (cfg = {}) => {
           region: process.env.VALKEY_AWS_REGION,
         },
       }
-      : process.env.VALKEY_PASSWORD ? {
-        username: process.env.VALKEY_USERNAME,
-        password: process.env.VALKEY_PASSWORD,
-      } : undefined
+      : process.env.VALKEY_AUTH_TYPE === "gcp-iam"
+        ? {
+          // "default" is the only supported username for GCP IAM authentication
+          // https://docs.cloud.google.com/memorystore/docs/valkey/manage-iam-auth#error-messages
+          username: "default",
+          password: await new GcpIAMProvider().getCredentials(),
+        }
+        : process.env.VALKEY_PASSWORD ? {
+          username: process.env.VALKEY_USERNAME,
+          password: process.env.VALKEY_PASSWORD,
+        } : undefined
 
   const useTLS = process.env.VALKEY_TLS === "true"
+  // Glide's TLS runs in its Rust core, so a custom CA must be passed explicitly
+  // via `rootCertificates` (Node's trust store / NODE_EXTRA_CA_CERTS do not apply).
+  const caCertPath = process.env.VALKEY_CA_CERT_PATH
+  const tlsAdvancedConfiguration = !useTLS
+    ? undefined
+    : process.env.VALKEY_VERIFY_CERT === "false"
+      ? { insecure: true }
+      : caCertPath
+        ? { rootCertificates: readFileSync(caCertPath) }
+        : undefined
   const sharedOptions = {
     addresses,
     credentials,
     useTLS,
     advancedConfiguration: {
-      ...(useTLS && process.env.VALKEY_VERIFY_CERT === "false" && {
-        tlsAdvancedConfiguration: {
-          insecure: true,
-        },
-      }),
+      ...(tlsAdvancedConfiguration && { tlsAdvancedConfiguration }),
       connectionTimeout: 30000,
     },
     requestTimeout: 5000,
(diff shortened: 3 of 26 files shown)


apps/server/src/iam-token-refresh.ts
```diff
@@ -0,0 +1,44 @@
...
+// Rotate the connection password for a gcp-iam client on a timer. 
+// Keyed by the client instance so shared cluster clients are only scheduled once; the timer
+// self-clears once the client is closed (updateConnectionPassword throws ClosingError),
+// so callers do not have to unregister at every close site.
+export function registerGcpTokenRefresh(client: RefreshableClient, label: string): void {
+  if (refreshTimers.has(client)) return
+
+  const timer = setInterval(async () => {
+    try {
+      const token = await mintGcpAccessToken()
```
nassery318: mintGcpAccessToken() is missing its required useTLS/verifyTlsCertificate args and fails to typecheck in local IDE. This throw at runtime before refreshing, so the token never rotates.

rezakaramad: @nassery318, It's fixed. `registerGcpTokenRefresh` now passes `useTLS` and `verifyTlsCertificate` through to `mintGcpAccessToken()` in all three spots (cluster, standalone, and orchestrator).

sepidehattar (approved):

apps/metrics/src/index.js
```diff
@@ -33,6 +34,40 @@ async function main() {
   const client = await createValkeyClient(cfg)
   const ownNodeId = sanitizeUrl(`${process.env.VALKEY_HOST}-${process.env.VALKEY_PORT}`)
 
+  // GCP OAuth2 tokens expire ~1h; rotate the connection password before then so
```
ArgusLi, line 37: Let's just reuse the common iam-token-refresh

apps/metrics/src/index.js
```diff
@@ -33,6 +34,40 @@ async function main() {
...
+  // reconnects keep authenticating. AWS IAM refreshes natively inside Glide.
+  let gcpRetryTimer
+  let gcpRetryUsedThisInterval = false
+  const refreshGcpToken = async () => {
+    try {
+      await client.updateConnectionPassword(await new GcpIAMProvider().getCredentials(), true)
```
ArgusLi: I believe there's no need to recreate the GcpIAMProvider here. Let's just save and reuse the object.

apps/metrics/src/index.js
```diff
@@ -33,6 +34,40 @@ async function main() {
...
+      // Success: drop any retry queued by an earlier failure.
+      if (gcpRetryTimer) {
```
ArgusLi: Let's pass in the timer too.

apps/metrics/src/valkey-client.js
```diff
@@ -33,23 +35,46 @@ export const createValkeyClient = async (cfg = {}) => {
...
+        : process.env.VALKEY_PASSWORD ? {
+          username: process.env.VALKEY_USERNAME,
+          password: process.env.VALKEY_PASSWORD,
+        } : undefined
 
   const useTLS = process.env.VALKEY_TLS === "true"
+  // The GCP IAM token is a bearer credential, so it must never travel over an
+  // unverified TLS channel. Refuse insecure TLS for gcp-iam regardless of
+  // VALKEY_VERIFY_CERT (mirrors the guard in GcpIAMProvider.getCredentials()).
+  const isGcpIam = process.env.VALKEY_AUTH_TYPE === "gcp-iam"
```
ArgusLi: Since this is only used once, let's just collapse it into the if statement.

common/src/gcp-iam-provider.ts
ArgusLi, line 1: Let's move this file to common and reuse in the metrics as well to maintain a single source of truth.

apps/server/src/iam-token-refresh.ts
ArgusLi, line 1: Let's move this to common and reuse for metrics as well.

apps/server/src/iam-token-refresh.ts
```diff
@@ -0,0 +1,98 @@
...
+// the token expires ~1h in, waiting a full interval could leave reconnects using
+// an expired token; retry sooner so a fresh token lands well before expiry.
+const RETRY_DELAY_MS = 5 * 60 * 1000
+
+type RefreshableClient = GlideClient | GlideClusterClient
+
+const refreshTimers = new Map<RefreshableClient, NodeJS.Timeout>()
+const retryTimers = new Map<RefreshableClient, NodeJS.Timeout>()
+// Clients whose current interval has already consumed its single retry.
+const retryUsed = new Set<RefreshableClient>()
```
ArgusLi: Instead of maintaining 2 maps and a set, let's just have 1 map. We can use have the map store an interface e.g.

```ts
interface RefreshState {
    label: string
    useTLS: boolean
    verifyTlsCertificate: boolean
    timer?: NodeJS.Timeout
    failures: number   // 0 while healthy; drives backoff while failing
  }
```

Then we can also utilize exponential backoff (`retryDelay` from `constants.ts`) instead of trying once and then waiting. Just need to make sure the backoff doesn't interfere with the regular interval.

ArgusLi (changes requested):
Thanks for this PR. A couple of things to reduce code duplication.

apps/server/src/iam-token-refresh.ts
rezakaramad: @ArgusLi Thanks for the review
This is done, single `Map<client, RefreshState>` with failures driving `retryDelay()` backoff. I used one self-rescheduling timer (success => 45 min, failure => backoff) so the backoff can't overlap the regular interval. Let me know if you'd prefer a fixed interval + separate backoff timer instead.

apps/server/src/iam-token-refresh.ts
rezakaramad: Moved to `iam-token-refresh.ts`; server + metrics now both use it.

common/src/gcp-iam-provider.ts
rezakaramad: Moved to `gcp-iam-provider.ts` as the single `mintGcpAccessToken`; deleted the server and metrics copies. Rhis adds `google-auth-library` to `common` deps.

apps/metrics/src/valkey-client.js
rezakaramad: Dropped the `isGcpIam` const and the standalone guard; `mintGcpAccessToken()` enforces the TLS + verification requirement.

apps/metrics/src/index.js
rezakaramad: This is done now. metrics now calls `registerGcpTokenRefresh` / `unregisterGcpTokenRefresh` from `valkey-common` and unregisters on shutdown.

apps/metrics/package.json
```diff
@@ -17,6 +17,7 @@
     "@smithy/signature-v4": "^5.3.13",
     "@valkey/valkey-glide": "^2.5.2",
     "express": "^4.21.2",
+    "google-auth-library": "^11.0.2",
```
ArgusLi: We can remove this as it's not referenced by any file in metrics anymore.

ArgusLi (commented):
This looks good, thank you for your quick changes.Just 1 final small change. 

Also could you please rebase the branch according to the [DCO action](https://github.com/valkey-io/valkey-admin/pull/508/checks?check_run_id=105093378132) and fix the lint?

apps/metrics/package.json
rezakaramad: @ArgusLi 
Appreciate your feedback - I guess we’re good to go now, if you don’t see anything else that needs addressing?

ArgusLi (approved):
Looks good to me. Thank you!

sepidehattar (approved):

rezakaramad:
@ArgusLi Hi, I’m not sure how we’ll handle the merge or what the flow will look like?
It would be nice if we could actually start using **valkey-admin** in GCP.
Thanks.

ArgusLi:
Hi @rezakaramad looks like some new changes were merged in main that conflicts with your branch. Just need to merge main into your branch and resolve the conflict.

We do appreciate the effort you've put in and would like to have memory store support in Valkey Admin. Once you've merged main in, I'll just ensure there are no bugs created on conflict resolution then re-approve and you can merge it in!

ArgusLi (approved):

ArgusLi:
@rezakaramad oh I think I understand now. Are you able to see or click `Squash and merge`? I can't see the repo permissions so can't tell if you can.

rezakaramad:
@ArgusLi Not really, I don't have `merge` rights apparently.

rezakaramad:
Wonderful, thanks so much @ArgusLi

ArgusLi:
I apologize, I thought you had merge rights once it was approved, that's why I approved and was waiting for you to merge it. It's a bit late for the 1.2.0 release, but it'll be included in the 1.3.0 release.

PR #512, opened 2026-09-09
Fix K8s sidecar registration and stale cluster topology broadcast

### LINKED ISSUE #510: K8s metrics sidecar registration rejected (401/404) in DEPLOYMENT_MODE=K8
## Summary
Metrics sidecars cannot register with the orchestrator in Kubernetes (`DEPLOYMENT_MODE=K8`). The sidecar retries and exits after 30 attempts, so metrics panels never populate.

```text
Register failed: 401 Unauthorized
Failed to register with server after 30 attempts. Shutting down.
```

## Root cause
Regressed in #379. The `/orchestrator/register` handler requires a pre-existing `metricsServerMap` entry for the nodeId. In K8s the orchestrator never spawns collectors (sidecars are externally managed), so no entry is ever pre-populated and every registration is rejected.

A secondary gap: sidecars have no per-node key to authenticate with, because the orchestrator only mints per-node keys for collectors it spawns (Web/Electron).

## Expected
A sidecar whose nodeId is part of the discovered cluster topology should be able to register (creating its entry on first register), authenticated by a shared key provisioned to both the orchestrator and the sidecars.

## Scope
- Register-time membership check gated to K8s; the strict no-entry gate is preserved for spawned collectors.
- Shared `ORCHESTRATOR_KEY` via a Kubernetes Secret; `resolveCollectorKey` falls back to it in K8s only.

### LINKED ISSUE #511: Cluster Topology view shows stale nodes: 30s refresh broadcasts but never re-discovers
## Summary
The Cluster Topology view reflects the topology captured at connect time and never updates. Nodes added or removed from the cluster afterward are not shown, even though a refresh loop runs every 30s.

## Root cause
Regressed in #279. `refreshAllClusterRegistries` (the 30s loop) broadcasts `clusterNodesRegistry` to clients but no longer re-discovers topology first — the per-cycle `updateClusterNodeRegistry` call was removed. So it re-broadcasts the same stale snapshot forever.

The frontend Cluster Topology route depends on this broadcast for node topology (mount only fetches INFO/utilization stats, not the node list), so there is no other path that refreshes it.

## Expected
The refresh loop should re-discover each tracked cluster's topology before broadcasting, so clients receive the current cluster nodes.

## Notes
Each cluster must be re-discovered through a live client that belongs to it (topology discovery keys off the client's own `CLUSTER SLOTS`); clusters with no live client are left unchanged.

ravjotbrar:
Fixes #510
Fixes #511

## What
Two coupled defects that broke the Kubernetes metrics path and left the Cluster Topology view stale.

### 1. K8s sidecar registration (#510)
- `/orchestrator/register` now allows a nodeId that belongs to the discovered cluster topology (in K8s), creating the sidecar's `metricsServerMap` entry on first register. The strict pre-existing-entry gate is preserved for spawned Web/Electron collectors.
- `resolveCollectorKey` falls back to a shared `ORCHESTRATOR_KEY` in K8s mode, since the orchestrator never spawns sidecars and so has no per-node key for them. The key is provisioned to both the orchestrator and the sidecars via a Kubernetes Secret.

### 2. Stale cluster topology broadcast (#511)
- `refreshAllClusterRegistries` now re-discovers each tracked cluster (through a live client belonging to that cluster) before broadcasting, so clients receive current topology instead of the connect-time snapshot. Clusters with no live client are left unchanged.

## Testing
- `npm test --workspace=server`: 246/246 pass (2 new).
- `npm run build --workspace=server`: success.
- `npx eslint` on changed files: clean.
- Live K8s verification to be done separately

### DIFF
--- a/apps/server/src/__tests__/metrics-orchestrator.test.ts
+++ b/apps/server/src/__tests__/metrics-orchestrator.test.ts
@@ -6,6 +6,7 @@ import {
   metricsServerMap,
   stopAllMetricsServers,
   reconcileClusterMetricsServers,
+  updateClusterNodeRegistry,
   clients,
   clusterNodesRegistry,
   __test__,
@@ -269,6 +270,77 @@ describe("metrics-orchestrator", () => {
       // metrics servers are started or stopped.
       assert.strictEqual(findDiff.mock.callCount(), 1)
       assert.strictEqual(updateMetricsServers.mock.callCount(), 0)
+    })
+  })
+
+  describe("topology refresh", () => {
+    afterEach(() => {
+      mock.restoreAll()
+      clusterNodesRegistry.clear()
+      metricsServerMap.clear()
+    })
+
+    it("updateClusterNodeRegistry replaces stale topology with the freshly discovered one", async () => {
+      // Seed a stale registry: cluster-1 knows only node1.
+      clusterNodesRegistry.set("cluster-1", {
+        node1: { host: "10.0.0.1", port: 6379, tls: false, verifyTlsCertificate: false },
+      })
+
+      // Discovery now returns an added node (node2) — i.e. the topology changed.
+      mock.method(__test__, "getClusterTopology", async () => ({
+        clusterId: "cluster-1",
+        discoveredClusterNodes: {
+          node1: { host: "10.0.0.1", port: 6379, tls: false, verifyTlsCertificate: false },
+          node2: { host: "10.0.0.2", port: 6379, tls: false, verifyTlsCertificate: false },
+        },
+      }))
+
+      await updateClusterNodeRegistry({} as never)
+
+      assert.deepStrictEqual(
+        Object.keys(clusterNodesRegistry.get("cluster-1") ?? {}).sort(),
+        ["node1", "node2"],
+        "registry should reflect the newly discovered topology, not the stale snapshot",
+      )
+    })
+
+    it("reconcile acts on refreshed topology: a node added by discovery is passed to updateMetricsServers", async () => {
+      mock.restoreAll()
+      clients.clear()
+
+      // Start with a registry + metrics map in sync on node1 only.
+      clusterNodesRegistry.set("cluster-1", {
+        node1: { host: "10.0.0.1", port: 6379, tls: false, verifyTlsCertificate: false },
+      })
+      metricsServerMap.set("node1", { metricsURI: "http://10.0.0.1:3000", pid: 123, lastSeen: Date.now() })
+
+      // Discovery reveals a new node2 → refresh the registry.
+      mock.method(__test__, "getClusterTopology", async () => ({
+        clusterId: "cluster-1",
+        discoveredClusterNodes: {
+          node1: { host: "10.0.0.1", port: 6379, tls: false, verifyTlsCertificate: false },
+          node2: { host: "10.0.0.2", port: 6379, tls: false, verifyTlsCertificate: false },
+        },
+      }))
+      await updateClusterNodeRegistry({} as never)
+
+      // findDiff derives adds/removes from whatever topology the registry now
+      // holds — so reconcile operates on the refreshed set, not the stale one.
+      mock.method(__test__, "findDiff", async (map: MetricsServerMap, nodes: ClusterNodeMap) => ({
+        nodesToAdd: Object.fromEntries(Object.entries(nodes).filter(([id]) => !map.has(id))),
+        nodesToRemove: [] as string[],
+      }))
+      const updateMetricsServers = mock.method(__test__, "updateMetricsServers", async () => {})
+
+      await reconcileClusterMetricsServers(metricsServerMap)
+
+      assert.strictEqual(updateMetricsServers.mock.callCount(), 1, "reconcile should act on the topology change")
+      const [nodesToAdd] = updateMetricsServers.mock.calls[0].arguments as [Record<string, unknown>, string[], string]
+      assert.deepStrictEqual(
+        Object.keys(nodesToAdd),
+        ["node2"],
+        "the node added by the refreshed topology should be reconciled",
+      )
     })
   })
 })
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -145,22 +145,8 @@ const wss = new WebSocketServer({ noServer: true })
 const delay = (ms: number) => new Promise((res) => setTimeout(res, ms))
 
 async function refreshAllClusterRegistries() {
-  // Re-discover each cluster's topology before broadcasting, so clients receive
-  // the current cluster nodes rather than a snapshot taken once at connect/startup.
-  // Each cluster is refreshed with a live client that belongs to it; a cluster we
-  // hold no client for is left as-is (nothing to rediscover through).
-  const clientByCluster = new Map<string, GlideClient | GlideClusterClient>()
-  for (const { client, clusterId } of clients.values()) {
-    if (clusterId && !clientByCluster.has(clusterId)) clientByCluster.set(clusterId, client)
-  }
-
-  await Promise.all(
-    [...clusterNodesRegistry.keys()]
-      .map((clusterId) => clientByCluster.get(clusterId))
-      .filter((client): client is GlideClient | GlideClusterClient => client != null)
-      .map((client) => updateClusterNodeRegistry(client)),
-  )
-
+  // Group live connections by cluster once; used both to pick a client to
+  // re-discover each cluster's topology and to target the broadcast below.
   const connectionIdsByCluster = new Map<string, string[]>()
   for (const [connectionId, entry] of clients) {
     if (!entry.clusterId) continue
@@ -168,6 +154,19 @@ async function refreshAllClusterRegistries() {
     ids.push(connectionId)
     connectionIdsByCluster.set(entry.clusterId, ids)
   }
+
+  // Re-discover each tracked cluster's topology before broadcasting, so clients
+  // receive the current cluster nodes rather than the connect-time snapshot.
+  // Each cluster is refreshed through one of its own live clients (topology
+  // discovery keys off that client's CLUSTER SLOTS); a cluster with no live
+  // client is left as-is, since there is nothing to rediscover through.
+  await Promise.all(
+    [...clusterNodesRegistry.keys()]
+      .map((clusterId) => connectionIdsByCluster.get(clusterId)?.[0])
+      .map((connectionId) => (connectionId ? clients.get(connectionId)?.client : undefined))
+      .filter((client): client is GlideClient | GlideClusterClient => client != null)
+      .map((client) => updateClusterNodeRegistry(client)),
+  )
 
   for (const [clusterId, clusterNodes] of clusterNodesRegistry) {
     const connectionIds = connectionIdsByCluster.get(clusterId)
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -3,6 +3,7 @@ import express from "express"
 import helmet from "helmet"
 import path from "path"
 import http from "http"
+import { GlideClient, GlideClusterClient } from "@valkey/valkey-glide"
 import {
   VALKEY,
   CONNECTION_TEARDOWN_DELAY_MS,
@@ -143,7 +144,23 @@ const wss = new WebSocketServer({ noServer: true })
 
 const delay = (ms: number) => new Promise((res) => setTimeout(res, ms))
 
-function refreshAllClusterRegistries() {
+async function refreshAllClusterRegistries() {
+  // Re-discover each cluster's topology before broadcasting, so clients receive
+  // the current cluster nodes rather than a snapshot taken once at connect/startup.
+  // Each cluster is refreshed with a live client that belongs to it; a cluster we
+  // hold no client for is left as-is (nothing to rediscover through).
+  const clientByCluster = new Map<string, GlideClient | GlideClusterClient>()
+  for (const { client, clusterId } of clients.values()) {
+    if (clusterId && !clientByCluster.has(clusterId)) clientByCluster.set(clusterId, client)
+  }
+
+  await Promise.all(
+    [...clusterNodesRegistry.keys()]
+      .map((clusterId) => clientByCluster.get(clusterId))
+      .filter((client): client is GlideClient | GlideClusterClient => client != null)
+      .map((client) => updateClusterNodeRegistry(client)),
+  )
+
   const connectionIdsByCluster = new Map<string, string[]>()
   for (const [connectionId, entry] of clients) {
     if (!entry.clusterId) continue
@@ -172,7 +189,7 @@ function refreshAllClusterRegistries() {
 async function refreshAllClusterRegistriesLoop() {
   while (true) {
     try {
-      refreshAllClusterRegistries()
+      await refreshAllClusterRegistries()
     } catch (err) {
       console.warn("Unable to refresh cluster topologies. ", err)
     } finally {
--- a/apps/server/src/metrics-orchestrator.ts
+++ b/apps/server/src/metrics-orchestrator.ts
@@ -79,7 +79,14 @@ const collectorKeys: Map<string, string> = new Map()
  * replayed against another.
  */
 export function resolveCollectorKey(nodeId: string): string | undefined {
-  return collectorKeys.get(nodeId)
+  const spawnedKey = collectorKeys.get(nodeId)
+  if (spawnedKey) return spawnedKey
+  // K8s collectors are external sidecars the orchestrator never spawns, so there
+  // is no per-node minted key. They authenticate with a shared key provisioned to
+  // both the orchestrator and every sidecar via a Kubernetes Secret. Scoped to K8s
+  // so the per-node property is preserved for spawned (Web/Electron) collectors.
+  if (isKubernetes) return process.env[ORCHESTRATOR_AUTH_KEY_ENV]
+  return undefined
 }
 
 /**
@@ -226,6 +233,18 @@ function isPinnedMetricsHost(uri: string): boolean {
 }
 
 /**
+ * True when `nodeId` is part of a discovered cluster topology. In K8s the
+ * orchestrator only tracks topology (it never spawns collectors), so this is
+ * how a sidecar's registration is authorized as belonging to the cluster.
+ */
+function isKnownClusterNode(nodeId: string): boolean {
+  for (const clusterNodes of clusterNodesRegistry.values()) {
+    if (flattenClusterNodeMap(clusterNodes)[nodeId]) return true
+  }
+  return false
+}
+
+/**
  * `POST /orchestrator/register` — a collector advertising where it can be
  * reached.
  *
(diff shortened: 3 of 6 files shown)


apps/server/src/metrics-orchestrator.ts
```diff
@@ -369,7 +396,7 @@ async function getClusterTopology(client: GlideClusterClient | GlideClient | nul
 
 export async function updateClusterNodeRegistry(client: GlideClusterClient | GlideClient | null, connectionDetails = initialConnectionDetails) {
   try {
-    const { discoveredClusterNodes, clusterId } = await getClusterTopology(client, connectionDetails)
+    const { discoveredClusterNodes, clusterId } = await internals.getClusterTopology(client, connectionDetails)
     if (clusterId && discoveredClusterNodes) clusterNodesRegistry.set(clusterId, discoveredClusterNodes)
```
alexey-temnikov, line 400: **This is the one item I'd resolve before merge.**

`refreshAllClusterRegistries` calls `updateClusterNodeRegistry(client)` with a single argument, so `connectionDetails` falls back to the env-derived `initialConnectionDetails`. `discoverCluster` builds every `NodeInfo` from those details, so each 30s cycle overwrites what was recorded at connect time.

Reproduced against this exact code path — connect with TLS + IAM, then make the refresh's own call:

```
at connect   : {"username":"real-cluster-user","tls":true,"authType":"iam","awsRegion":"us-east-1"}
after refresh: {"username":"preconfigured-user","tls":false}
```

`tls` is downgraded to `false`, IAM `authType`/`awsRegion` are dropped, and `username` is replaced with a *different* cluster's env value. Since `clusterNodesRegistry` feeds `startMetricsServers`, a Web deployment with a preconfigured connection (where `runReconcileLoop` is active) will spawn collectors for newly added nodes using the wrong auth material.

The narrower fix is to stop letting a detail-less refresh clobber per-node auth fields:

```suggestion
    const { discoveredClusterNodes, clusterId } = await internals.getClusterTopology(client, connectionDetails)
    if (clusterId && discoveredClusterNodes) {
      // A refresh re-discovers through a live client and passes no connectionDetails,
      // so `discoverCluster` would stamp env-derived defaults over the real per-node
      // auth metadata. Keep what was already recorded for nodes we already knew.
      const existing = clusterNodesRegistry.get(clusterId)
      clusterNodesRegistry.set(clusterId, Object.fromEntries(
        Object.entries(discoveredClusterNodes).map(([nodeId, node]) => {
          const prev = existing?.[nodeId]
          return [nodeId, prev
            ? {
              ...node,
              username: prev.username,
              tls: prev.tls,
              verifyTlsCertificate: prev.verifyTlsCertificate,
              authType: prev.authType,
              awsRegion: prev.awsRegion,
              awsReplicationGroupId: prev.awsReplicationGroupId,
            }
            : node]
        }),
      ))
    }
```

The cleaner alternative is to record the `ConnectionDetails` per `clusterId` at connect time (alongside the existing `clusterCredentials`) and pass them on refresh, so discovery always runs with the details that actually opened the connection. Either way, worth a test asserting a refreshed entry keeps `tls: true` and its `authType`.

apps/server/src/index.ts
```diff
@@ -152,6 +155,19 @@ function refreshAllClusterRegistries() {
...
+  // Each cluster is refreshed through one of its own live clients (topology
+  // discovery keys off that client's CLUSTER SLOTS); a cluster with no live
+  // client is left as-is, since there is nothing to rediscover through.
+  await Promise.all(
+    [...clusterNodesRegistry.keys()]
+      .map((clusterId) => connectionIdsByCluster.get(clusterId)?.[0])
+      .map((connectionId) => (connectionId ? clients.get(connectionId)?.client : undefined))
+      .filter((client): client is GlideClient | GlideClusterClient => client != null)
+      .map((client) => updateClusterNodeRegistry(client)),
+  )
```
alexey-temnikov, line 169: The awaited `Promise.all` has no deadline. I confirmed a hung `CLUSTER SLOTS` leaves this call unsettled indefinitely, which stalls the whole loop — so one unresponsive node stops the broadcast for *every* cluster, not just its own. Before this PR the function was synchronous and couldn't stall this way.

`delay` is already defined just above, so bounding it is cheap:

```suggestion
  // Re-discover each tracked cluster's topology before broadcasting, so clients
  // receive the current cluster nodes rather than the connect-time snapshot.
  // Each cluster is refreshed through one of its own live clients (topology
  // discovery keys off that client's CLUSTER SLOTS); a cluster with no live
  // client is left as-is, since there is nothing to rediscover through.
  // Each re-discovery is bounded: this is awaited by the refresh loop, so an
  // unsettled discovery would block every cluster's broadcast, not just its own.
  await Promise.all(
    [...clusterNodesRegistry.keys()]
      .map((clusterId) => connectionIdsByCluster.get(clusterId)?.[0])
      .map((connectionId) => (connectionId ? clients.get(connectionId)?.client : undefined))
      .filter((client): client is GlideClient | GlideClusterClient => client != null)
      .map((client) => Promise.race([
        updateClusterNodeRegistry(client),
        delay(TOPOLOGY_REDISCOVERY_TIMEOUT_MS).then(() => {
          console.warn("Topology re-discovery timed out; broadcasting last known nodes.")
        }),
      ])),
  )
```

(with `const TOPOLOGY_REDISCOVERY_TIMEOUT_MS = 10_000` alongside the other module constants).

apps/server/src/index.ts
```diff
@@ -143,7 +144,9 @@ const wss = new WebSocketServer({ noServer: true })
 
 const delay = (ms: number) => new Promise((res) => setTimeout(res, ms))
 
-function refreshAllClusterRegistries() {
+async function refreshAllClusterRegistries() {
```
alexey-temnikov, line 147: **#510 is still broken for a headless Kubernetes deployment — worth covering before this closes the issue.**

`isKnownClusterNode` reads `clusterNodesRegistry`, which in K8s is populated once by `updateRegistryforK8()`. This new re-discovery only refreshes clusters that have a live entry in `clients`, and `getInitialClient()` never adds itself there — so `clients` stays empty until a browser connects.

I hit this on real Kubernetes. Starting from a working 6-node cluster (all sidecars registered), I scaled the StatefulSet to 8 and joined both new nodes properly (`cluster_state:ok`, new primary owning slots). With no UI session open:

```
Rejected metrics registration for valkey-6-...-6379: no metrics server entry
Rejected metrics registration for valkey-7-...-6379: no metrics server entry
```

…repeated 180 times. The instant I opened a WebSocket session, the refresh ran and both registered:

```
Metrics server registered for valkey-6-...-6379 at http://10.42.0.32:3000
Metrics server registered for valkey-7-...-6379 at http://10.42.0.31:3000
```

So on a headless deployment — the normal steady state for a monitoring backend — a scaled-out node's sidecar never registers, and it recovers only by chance when someone opens the UI. Since the K8s path already has its own entry point, something like this would close the gap:

```ts
// in the K8s branch, independent of `clients`
if (isKubernetes) {
  const client = await getInitialClient()
  await updateClusterNodeRegistry(client, initialConnectionDetails)
}
```

Related: `updateRegistryforK8()` currently floats its promise (`updateClusterNodeRegistry(...)` is not awaited), so a slow first discovery races the sidecars' 30-attempt/30-second budget.

apps/server/src/__tests__/metrics-orchestrator.test.ts
```diff
@@ -271,4 +272,75 @@ describe("metrics-orchestrator", () => {
...
+      }))
+      await updateClusterNodeRegistry({} as never)
+
+      // findDiff derives adds/removes from whatever topology the registry now
+      // holds — so reconcile operates on the refreshed set, not the stale one.
+      mock.method(__test__, "findDiff", async (map: MetricsServerMap, nodes: ClusterNodeMap) => ({
+        nodesToAdd: Object.fromEntries(Object.entries(nodes).filter(([id]) => !map.has(id))),
+        nodesToRemove: [] as string[],
+      }))
+      const updateMetricsServers = mock.method(__test__, "updateMetricsServers", async () => {})
```
alexey-temnikov, line 333: This mock makes the test unable to fail. It replaces `findDiff` with a reimplementation of the very logic the assertions then check, so the test verifies its own mock rather than production code.

I confirmed it empirically: I replaced the real `findDiff` body so it always returns no additions, and **both** new "topology refresh" tests still passed — while the pre-existing `findDiff` test correctly failed. Letting the real `findDiff` run makes the test meaningful, and it passes as written (`node1` stays, `node2` is added):

```suggestion
      // Let the real findDiff derive adds/removes from the refreshed registry, so
      // this fails if reconcile stops acting on the new topology.
      const updateMetricsServers = mock.method(__test__, "updateMetricsServers", async () => {})
```

Both `MetricsServerMap` and `ClusterNodeMap` are used elsewhere in the file, so the imports stay live.

Bigger picture: the actual #511 fix lives in `refreshAllClusterRegistries` in `index.ts` and has no coverage here — both new tests exercise `updateClusterNodeRegistry`/`reconcile`, which were already correct. Neither would catch the `connectionDetails` regression flagged in `metrics-orchestrator.ts`. Note too that `isKubernetes` is a module-load constant, so none of the new K8s branches (`resolveCollectorKey` fallback, `isKnownClusterNode`, the `isPinnedMetricsHost` bypass) are reachable from unit tests as written.

examples/k8s/valkey-statefulset-sidecar-patch.yaml
```diff
@@ -34,6 +34,11 @@ spec:
                   fieldPath: status.podIP
             - name: METRICS_ADVERTISE_PORT
               value: "3000"
+            - name: ORCHESTRATOR_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: valkey-admin-orchestrator-key
+                  key: ORCHESTRATOR_KEY
```
alexey-temnikov, line 41: `examples/k8s/valkey-statefulset.yaml` ships this same metrics sidecar inline but didn't get the `ORCHESTRATOR_KEY` block, so it isn't part of this diff and I can't suggest into it directly. It currently has zero occurrences of `ORCHESTRATOR_KEY` (vs 2 here and 4 in `app.yaml`), so anyone using that sample gets a keyless sidecar that exits immediately with `Missing ORCHESTRATOR_KEY ... Shutting down.` The same five lines are needed on its `metrics` container:

```yaml
            - name: ORCHESTRATOR_KEY
              valueFrom:
                secretKeyRef:
                  name: valkey-admin-orchestrator-key
                  key: ORCHESTRATOR_KEY
```

While in there: that manifest starts `--cluster-enabled yes` nodes but never forms the cluster (no `CLUSTER MEET`/slot assignment), so `CLUSTER SLOTS` comes back empty and the sidecar dies on `Connection error: Cluster(No topology views found)`. Because StatefulSet readiness is ordered, that crash-loop blocks the whole rollout at `valkey-0` — I had to write a cluster-forming manifest to test this PR at all. Either form the cluster in that sample or note that it needs `valkey-cli --cluster create` before the sidecars can start.

docs-site/src/content/docs/deployment/kubernetes.md
```diff
@@ -181,15 +196,21 @@ kubectl logs -n valkey valkey-0 -c metrics
...
 
 ```text
 Register failed: 401 Unauthorized
 Failed to register with server after 30 attempts. Shutting down.
 ```
 
-Support  is in progress. Until then the Kubernetes sidecar topology on this page will deploy but will not populate metrics panels. Use the [Docker deployment](/deployment/docker/) for a working metrics path.
+If you see this, confirm the `valkey-admin-orchestrator-key` Secret exists in the
+`valkey` namespace and that both pods reference it (`kubectl get pod ... -o yaml |
+grep -A3 ORCHESTRATOR_KEY`).
```
alexey-temnikov, line 213: Two properties of the shared key are worth stating here, since this PR is what makes them reachable. I probed the boundary with signed requests against a live deployment: the primitive itself holds up (wrong key → 401 `invalid_signature`, stale timestamp → 401 `stale_timestamp`, node outside the topology → 401). But:

- the key is cluster-wide, so any single compromised sidecar can register as **any** node — I registered as `valkey-3` while not being it and got `200 Registered node`;
- `isPinnedMetricsHost` returns `true` unconditionally when `isKubernetes`, so an arbitrary `metricsServerUri` is accepted — both `http://169.254.169.254/` (IMDS) and `http://attacker.example:9999` returned `200`.

The host-pin bypass predates this PR, but combined with the shared key it becomes reachable by one compromised sidecar, and the server does fetch from whatever URI is registered. Suggesting a short operator-facing note:

```suggestion
If you see this, confirm the `valkey-admin-orchestrator-key` Secret exists in the
`valkey` namespace and that both pods reference it (`kubectl get pod ... -o yaml |
grep -A3 ORCHESTRATOR_KEY`).

The key is shared cluster-wide rather than per-node, so it authenticates a sidecar
as *a* member of the cluster, not as one specific node. Treat it as a
cluster-scoped credential: restrict read access on the Secret, rotate it by
recreating the Secret and restarting the app Deployment and the sidecars, and note
that in `DEPLOYMENT_MODE=K8` the registered metrics host is not pinned to
loopback, so any holder of the key can point a node's metrics URI at an arbitrary
address.
```

apps/server/src/metrics-orchestrator.ts
```diff
@@ -225,6 +232,18 @@ function isPinnedMetricsHost(uri: string): boolean {
...
+ * True when `nodeId` is part of a discovered cluster topology. In K8s the
+ * orchestrator only tracks topology (it never spawns collectors), so this is
+ * how a sidecar's registration is authorized as belonging to the cluster.
+ */
+function isKnownClusterNode(nodeId: string): boolean {
+  for (const clusterNodes of clusterNodesRegistry.values()) {
+    if (flattenClusterNodeMap(clusterNodes)[nodeId]) return true
+  }
+  return false
+}
```
alexey-temnikov, line 245: `isKnownClusterNode` scans every entry in `clusterNodesRegistry`, and that map can accumulate orphans, because `clusterId` is derived as `CLUSTER SLOTS[0][2][2]` — the node id of whoever owns slot 0. Node ids are stable per node, so a failover or slot migration changes the derived `clusterId` and re-discovery writes a **new** key while the old snapshot stays behind.

Observed live: the app started when `valkey-0` (`d44f6f20…`) owned slot 0; after a reshard `valkey-6` (`e86611e2…`) owned it, leaving two entries for one cluster. The orphan is never broadcast (`clients[].clusterId` is set at connect time and never updated), but it persists for the process lifetime and still feeds this function — so a node that has since been removed from the cluster keeps its registration rights indefinitely.

The derivation predates this PR, but the 30s re-discovery is what makes it happen at runtime. Options: key the registry on something stable (`CLUSTER MYID` of the connected node, or the connection's own identity), or have the refresh replace the entry it was asked to refresh rather than inserting under a freshly derived id.

apps/server/src/index.ts
ArgusLi: Yup this makes sense, I will say it makes sense to add which client it is to the console warn if possible.

apps/server/src/metrics-orchestrator.ts
```diff
@@ -269,14 +288,22 @@ function handleRegister(req: Request, res: Response): void {
...
+  const allowed = metricsServerMap.has(nodeId) || (isKubernetes && isKnownClusterNode(nodeId))
+  if (!allowed) {
     console.warn(`Rejected metrics registration for ${nodeId}: no metrics server entry`)
     res.status(401).send("Unauthorized")
     return
   }
 
+  let entry = metricsServerMap.get(nodeId)
+  if (!entry) {
+    entry = { metricsURI: "", pid: undefined, lastSeen: Date.now() }
```
ArgusLi, line 308: Shouldn't the metricsURI here be metricsServerUri?

apps/server/src/metrics-orchestrator.ts
ArgusLi: We should revisit this. This should be out of scope for this fix, but is a very real, valid issue. There's actually no permanent Valkey issued cluster id.

apps/server/src/index.ts
ArgusLi: Should we even enable headless deployment? This issue stems with allowing users to set an initial client using environment variables. This means that without opening the UI, the app can connect to a Valkey cluster. But what features does the app provide in this state? While it can collect historical data, there is no way to consume the data collected without ever opening the UI.

In addition, this adds complexity to the design. Clients is session-scoped based on websocket connections. Once there are no more websockets (UI instances) watching the client, it gets torn down by the session-expiry listener.

Using this fix makes the initialClient instance last forever and when torn down, self-healing. My suggestion is to completely remove initialClient.

ArgusLi: Adding on, the initial client is not even present in any UI connecting to the server. Even if a user connects in the UI with the same connection information, it registers a new client, overwriting the initialClient.

apps/server/src/metrics-orchestrator.ts
ArgusLi: We should remove making connectionDetails optional (and defaulting to initialConnection) in updateClusterNodeRegistry(). The function should be straightforwardly using the passed connectionDetails to conduct the operation, and doing so would prevent errors like this.

We should create a new function to check for existing connectionDetails, and use it to pass them to updateClusterNodeRegistry().

ArgusLi (changes requested):

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Sorry, I'm a bit confused here. How does a failed node keep its registration rights forever?

docs-site/src/content/docs/deployment/kubernetes.md
ravjotbrar: I agree documentation is a good way to communicate this to the user.

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Agree with both points.

apps/server/src/index.ts
ravjotbrar: I disagree with removing initial client. This is an ask from users to be able to begin metrics collection without manually connecting to an instance through the UI. There are even issues that want to enhance the ability to preconfigure multiple clusters, rather than just the one.

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Changed.

apps/server/src/metrics-orchestrator.ts
ravjotbrar: Looked into this a bit more closely and I understand now. updateClusterRegistry never should've been re-deriving the clusterId. I'll add it as a param to the method and pass in the known Id.

apps/server/src/metrics-orchestrator.ts
ravjotbrar: The metricsServerMap uses metricsURI

apps/server/src/__tests__/metrics-orchestrator.test.ts
ravjotbrar: K8s branches now reachable — added orchestrator-k8s-register.test.ts, which sets DEPLOYMENT_MODE=K8 before import (no refactor). Covers resolveCollectorKey's shared-key fallback and handleRegister admitting a signed sidecar for a known-topology node with no pre-existing entry (+ rejecting unknown nodes / wrong keys). - connectionDetails regression covered — extracted resolveClusterRefreshTarget and added a test that a user connected cluster carries its own node metadata forward instead of reverting to initialConnectionDetails; a preconfigured-mode test covers the headless fallback.

alexey-temnikov (dismissed):

ArgusLi (approved):

PR #522, opened 2026-09-18
Fail closed instead of storing a cleartext password when no OS keystore is available

ravjotbrar:
## Problem
On systems with no OS secure store, `safeStorage.isEncryptionAvailable()` is false and the encrypt IPC returned the **plaintext** password. The renderer then persisted that plaintext to `localStorage` while the UI implied it was encrypted.

## Fix (fail closed)
- **Encrypt IPC** (`electron.main.js`) returns a structured `{ ok, value }` and **never** returns plaintext; added a `secure-storage:is-encryption-available` IPC. `preload.js` + `secureStorage.ts` updated (`encryptForStorage`, `isEncryptionAvailable`; removed the plaintext-returning path).
- **Connect/save paths** keep the entered plaintext in memory so the current session still connects, mark the connection `isPasswordEncrypted = false`, and `toast.warning(...)` the user with an actionable keyring hint. Applied at every password-entry site (new connection, edit, password prompt, reconnect, cluster node), including threading the flag through cluster discovery.
- **`persistConnections`** is now the single localStorage write path: a connection whose password could not be encrypted is written with `password: undefined` (and the transient flag dropped), so **no cleartext reaches disk**. On next use the connection re-prompts (existing `R.isNil(password)` prompt path).

## Behavior
- Encryption available → unchanged: ciphertext stored, connects, no prompt on reload.
- Encryption unavailable → connects this session, user warned, nothing persisted, re-prompts next time.
- **Passwordless** connections are unaffected (encryption is never invoked).


apps/frontend/src/state/epics/valkeyEpics.ts
```diff
@@ -61,7 +79,7 @@ export const connectionEpic = (store: Store) =>
         if (R.isNil(password) || action.payload.connectionDetails.authType === "iam") return action
         
         // Password is dispatched as plaintext if secureStorage is unavailable
-        const decryptedPassword = password.length > 0 && secureStorage.isAvailable() ? await secureStorage.decrypt(password) : password
+        const decryptedPassword = password.length > 0 && secureStorage.isElectron() ? await secureStorage.decrypt(password) : password
```
ArgusLi: Why this change? We cannot assume secureStorage is available if isElectron.

apps/frontend/src/utils/secureStorage.ts
```diff
@@ -1,31 +1,45 @@
...
-    return password
+  // True only when the OS actually has a secure store (not merely that we are in
+  // Electron). Reflects safeStorage.isEncryptionAvailable() in the main process.
+  isEncryptionAvailable: async (): Promise<boolean> => {
+    if (!window.secureStorage?.isEncryptionAvailable) return false
+    return await window.secureStorage.isEncryptionAvailable()
   },
 
-  isAvailable: (): boolean => {
+  isElectron: (): boolean => {
```
ArgusLi, line 31: Is there ever a case in the code where only isElectron is needed instead of if encryption is needed? I think we can remove isElectron entirely and replace with isEncryptionAvailable.

apps/frontend/src/components/cluster-topology/cluster-node-row.tsx
```diff
@@ -126,14 +129,16 @@ export function ClusterNodeRow({
   }
 
   const handlePasswordSubmit = async (password: string) => {
-    const encryptedPw = await secureStorage.encryptIfAvailable(password)
+    const result = await secureStorage.encryptForStorage(password)
+    if (!result.ok && secureStorage.isElectron()) toast.warning(PASSWORD_NOT_STORED_WARNING)
```
nassery318: Since the content of the Toast are long - we should give the user enough time to be able to read it. By default Sonner has 4 seconds and it is not enough. so we can add the duration to it.

toast.warning(PASSWORD_NOT_STORED_WARNING, { duration: 15000 })

apps/frontend/src/components/connection/Connection.tsx
```diff
@@ -49,10 +50,12 @@ export function Connection() {
     if (!passwordPromptConnectionId) return
     const connection = connections[passwordPromptConnectionId]
     if (!connection) return
-    const encryptedPassword = await secureStorage.encryptIfAvailable(password)
+    const result = await secureStorage.encryptForStorage(password)
+    if (!result.ok && secureStorage.isElectron()) toast.warning(PASSWORD_NOT_STORED_WARNING)
```
nassery318: Same here.
toast.warning(PASSWORD_NOT_STORED_WARNING, { duration: 15000 })

apps/frontend/src/components/ui/connection-form.tsx
```diff
@@ -87,22 +88,29 @@ function ConnectionForm({ onClose }: ConnectionFormProps) {
...
-      : trimmed
+    let isPasswordEncrypted: boolean | undefined
+    let detailsToDispatch = trimmed
+    if (connectionDetails.password) {
+      const result = await secureStorage.encryptForStorage(connectionDetails.password)
+      isPasswordEncrypted = result.ok
+      // On failure keep the plaintext so this session can connect; it will not be
+      // persisted (see the persistence layer), and the user is warned.
+      detailsToDispatch = { ...trimmed, password: result.ok ? result.value : connectionDetails.password }
+      if (!result.ok && secureStorage.isElectron()) toast.warning(PASSWORD_NOT_STORED_WARNING)
```
nassery318: Same here.
toast.warning(PASSWORD_NOT_STORED_WARNING, { duration: 15000 })

apps/frontend/src/components/ui/edit-form.tsx
```diff
@@ -146,15 +147,23 @@ function EditForm({ onClose, connectionId }: EditFormProps) {
...
-        : trimmed
+      // and carries the source connection's marking (the connect below targets a new
+      // connectionId, so the reducer can't inherit it).
+      let isPasswordEncrypted = passwordChanged ? undefined : fullConnection?.isPasswordEncrypted
+      let detailsToDispatch = trimmed
+      if (passwordChanged && connectionDetails.password) {
+        const result = await secureStorage.encryptForStorage(connectionDetails.password)
+        isPasswordEncrypted = result.ok
+        detailsToDispatch = { ...trimmed, password: result.ok ? result.value : connectionDetails.password }
+        if (!result.ok && secureStorage.isElectron()) toast.warning(PASSWORD_NOT_STORED_WARNING)
```
nassery318: Same in here.
toast.warning(PASSWORD_NOT_STORED_WARNING, { duration: 15000 })

apps/frontend/src/components/ValkeyReconnect.tsx
```diff
@@ -51,10 +52,12 @@ export function ValkeyReconnect() {
 
   const handlePasswordSubmit = async (password: string) => {
     if (!connection) return
-    const encryptedPassword = await secureStorage.encryptIfAvailable(password)
+    const result = await secureStorage.encryptForStorage(password)
+    if (!result.ok && secureStorage.isElectron()) toast.warning(PASSWORD_NOT_STORED_WARNING)
```
nassery318: same.
toast.warning(PASSWORD_NOT_STORED_WARNING, { duration: 15000 })

apps/frontend/src/utils/secureStorage.ts
```diff
@@ -1,31 +1,45 @@
...
+
 export const secureStorage = {
-  encrypt: async (unencrypted: string): Promise<string> => {
-    if (!unencrypted || !window.secureStorage) return ""
-    return await window.secureStorage.encrypt(unencrypted)
+  // Encrypt for PERSISTENCE. Reports whether real encryption happened so the
+  // caller can refuse to persist an unprotected secret.
+  encryptForStorage: async (password: string): Promise<EncryptResult> => {
+    if (!password) return { ok: true, value: "" }
+    if (!window.secureStorage) return { ok: false }
```
ArgusLi, line 15: Instead of checking secureStorage we should check for isEncryptionAvailable. Having secureStorage doesn't guarantee encryption is available.

apps/frontend/src/utils/secureStorage.ts
```diff
@@ -1,31 +1,45 @@
+type EncryptResult = { ok: true; value: string } | { ok: false }
+
+// Electron-only: shown when the bridge exists but the OS has no secure store
+// (gate on secureStorage.isElectron()). Web has no secure store by design and
+// never persists real passwords, so it must not warn.
+export const PASSWORD_NOT_STORED_WARNING =
+  "This system has no secure credential store, so the password can't be saved and will be requested " +
+  "on the next connection. Install or unlock your system keyring (e.g. gnome-keyring or KWallet) to " +
```
ArgusLi: nit: this displays on non linux too. My suggestion is to directly remove the linux specific system keyring examples.

ravjotbrar: Which non linux do we support that doesn't have secure storage available?

apps/frontend/src/utils/secureStorage.ts
ravjotbrar: The isElectron gates against showing the warning on Web

apps/frontend/src/utils/secureStorage.ts
ravjotbrar: you never need isEncrpytionAvailable directly, because encryptForStorage already tells you. When you try to encrypt, the main process checks availability itself and returns either the ciphertext (ok: true) or a refusal (ok: false). So the answer to "is encryption available?" comes back with every encrypt attempt.

apps/frontend/src/utils/secureStorage.ts
ArgusLi: It's possible for an OS to have secure storage support but it's not available, e.g. MacOS without Keychain.

This warning displays when `!result.ok && secureStorage.isElectron()`. If the Electron app is running on MacOS without Keychain access, it'll return result.ok false as `safeStorage.isEncryptionAvailable()` will be false on encryption and so the warning will display.

ArgusLi (dismissed):

ArgusLi (approved):

Issue #519, opened 2026-09-16
Cannot switch databases in Valkey Admin after connecting


masterwishx:
### Before submitting

- [x] I searched existing issues and didn't find a duplicate

### Description
Valkey Admin connects only to the database number specified in the connection settings.  
There is currently no option in the UI to switch to another database (e.g., DB 1, DB 2, etc.) after the connection is established.

### Expected Behavior
A database selector option should allow switching between databases without editing the connection settings.

### Actual Behavior
The UI provides no way to change the active database.  
The only method is to edit the connection settings and reconnect, which is inconvenient and breaks workflows that require browsing multiple DBs.

### Steps to Reproduce
1. Open Valkey Admin.
2. Create a connection with a specific DB number (e.g., 0) or 15.
3. Connect to the server.
4. Attempt to switch to DB 1 or any other DB.

### Result
There is no UI element to switch databases.

### Deployment
- Valkey Admin version: latest
- Web (Docker)
- OS: Unriad

### Additional Notes
Redis/Valkey supports multiple databases via the `SELECT <db>` command, but Valkey Admin does not expose this functionality in UI.
So there is no option to switch db in UI. 

`SELECT 1` in command is working but we can see changes only in Key browser section while still connected to db 0 : 
```
Key Browser
Add, View and Edit keys of instance 192-168-0-199-6379-db0
```

But here we connected to db1 (after `SELECT 1`)

### Screenshots
<img width="1491" height="947" alt="Image" src="https://github.com/user-attachments/assets/68a70ee9-2fa9-44a8-a86e-c511eece7491" />

nassery318:
Thanks @masterwishx for creating this issue. Please see below our findings as we address this issue:

- This is **by design**. Each `(host, port, db)` is treated as its own connection, so switching databases means adding a separate connection for that DB.
- The problem is that this behavior **isn't obvious to the user**.
- As a solution, we could add a **database dropdown** in the app to make switching to or opening another database easier.

masterwishx:
> * As a solution, we could add a **database dropdown** in the app to make switching to or opening another database easier.

Thanks, that will be helpful.

ArgusLi:
While not the hardest issue, I would not consider it a good first issue.

Adding a UI element to change DB would require coordination with the server to:
1. Check if there's an existing client for the new connection (DB).
    a. If it exists, use that client.
    b. If it doesn't exist, create a new client.
2. Clean up the current client if needed.

PR #524, opened 2026-09-18
Require confirmation for SELECT on cluster connections

### LINKED ISSUE #462: SELECT command in Send Command can change database on shared cluster client
## Description

On cluster connections, all connectionIds share a single `GlideClusterClient`. If a user runs `SELECT <N>` via Send Command, it changes the active database on the shared connection, affecting all other users/connections using that cluster client.

## Possible Fix

Add `SELECT` to the command confirm list for cluster connections, since database selection should only happen at connection time via the `databaseId` client option.

bharathsr08:
Running SELECT in Send Command can change the selected database on a shared cluster client and affect other operations. Require confirmation for SELECT when the console is opened in cluster context, while preserving standalone behavior and existing confirmation rules.

Add nine regression cases covering cluster context, case-insensitive and quoted commands, omitted context, existing rules, and unrelated commands. Run the common test suite in CI and document the warning in the README and Send Command guide.

Fixes #462.

Validation:
- All 84 common tests passed, including 22 command-restriction tests.
- ESLint passed for all changed TypeScript files.
- Shared package and frontend production builds passed.
- Git diff whitespace check passed.

Distribution validation is pending: dispatch attempts for Docker, Linux, and macOS workflows on the fork returned HTTP 404. This PR remains a draft until the distribution checks required by CONTRIBUTING.md can run successfully.


ArgusLi:
Hi @bharathsr08, thank you for your PR submission. While I understand you applied the potential fix listed in #462  and I appreciate the work you've done here, I believe allowing SELECT in send command should not be allowed as it changes the state of the Valkey Client for other users using that client. Instead, we should prevent users from changing client state in Send command, with the goal of moving all that functionality to the UI instead.

I'll update the issue accordingly.

bharathsr08:
Thanks @ArgusLi for looking at this.

I’d like to take on this change. I propose extending the existing shared blocked-command list, enforced in both the frontend and backend, to reject connection-state changes on standalone and cluster connections. 
Database selection would remain in the connection UI. 

I'll make an initial change to track the restricted commands and possibly a later one to move the functionality to UI. My proposed initial scope is SELECT/AUTH/HELLO/RESET/QUIT. Does this sound ok?

ArgusLi:
Hi @bharathsr08 that sounds good to me, thanks! Please tag #462 in your PR as you've done in this one.

Note adding DB selection in the connection UI is tracked by #519 and is more complex than you are describing.

